Issue #307, opened 2014-12-02
systemd roscore.service use wrong environment


simkim:
When installing roslaunch-systemd from roslaunch recipe, the systemd configuration don't seems to match installed environnement.
- It use variable in variable for systemd which is not supported ( http://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment= )
- It miss PYTHONPATH, PATH, LD_LIBRARY_PATH variables
- It use /usr/bin/roscore instead of /opt/ros/hydro/bin/roscore

bulwahn:
Now, as the pull request #308 has been merged, this issue is resolved. @simkim Can you confirm this and close the issue?

bulwahn:
As I believe that the issue has been resolved, I close it now.

PR #308, opened 2014-12-02
roslaunch: Use as documented environment

simkim:
in relation to #307 

roscore-default: Removing not supported variable in variable
roscore-default: Adding PATH, LD_LIBRARY_PATH and PYTHONPATH
roscore.service: Using /opt/ros/hydro/bin instead of /usr/bin to launch
roscore

You can now install roslaunch-systemd :

opkg install roslaunch-systemd

and start roscore :

systemctl start roscore


bulwahn:
:+1:

PR #310, opened 2014-12-24
festival: correcting audsp path (resolves #309)

### LINKED ISSUE #309: festival: failed to start audsp
Festival is not working correctly:

```
# echo "test" | festival --tts
pipe_open: failed to start audsp
Audio spooler has died unexpectedly
```

Reason: audsp is not found, as it is not in the PATH, and not executable. It is shipped to `/usr/lib/audsp`.

An easy hack is to put it in `/usr/bin`, but it seems that on Ubuntu it is placed in `/usr/lib/festival` instead.

KristofRobot:


bulwahn:
:+1:

PR #311, opened 2015-01-09
Generalize construction of ros_prefix and use it consistently

elmi82:
The ros_prefix currently handled special situations like building a native package with a virtclass override. This can be handled more easily by basing the prefix on yocto's base prefix. This also sets the prefix correctly if one wants to build nativesdk packages.

Additionally the special ros_prefixes should be use in all ros packages to ensure the resulting packages contain all necessary assets.


bulwahn:
:+1:

PR #319, opened 2015-03-03
 Reduction of runtime dependencies

andreasbaak:
The packages roslaunch and roslib pull in coreutils, gcc, cmake and
other build utilties as runtime dependencies that are all deployed on
the target board together with the ros packages. Most of the
dependencies are not needed at runtime. Drawbacks in the current state:
-  increased the image size (gcc is a big package)
-  coreutils is licensed under GPLv3 and 
  must not be deployed in some production systems

Therefore, the two patches remove the critical runtime dependencies.


bulwahn:
I have already discussed the changes with Andreas in person. So, :+1: from my side.

PR #327, opened 2015-05-28
Update from virtclass-native to class-native

ashcharles:
As per Section 24.5.15 of the Yocto Manual [1], use of 'virtclass'
overrides has been deprecated since Yocto version 1.6. Update to
the new syntax.

[1] http://www.yoctoproject.org/docs/1.8/mega-manual/mega-manual.html

Signed-off-by: Ash Charles ashcharles@gmail.com


bulwahn:
:+1: Looks good.

PR #358, opened 2015-10-05
Added gst-plugins-good to DEPENDS line

wintermuteger:
Some of the gst-plugins-good modules were missing to setup a cam for some basic scenarios/tutorials. I added it so they'll be added to the default build set.


bulwahn:
:+1:

PR #359, opened 2015-10-06
Added rosbridge_suite (incl. rosbridge_server) to meta-ros

wintermuteger:
Changes to be committed:  
        new file:   rosapi_0.7.13.bb
    new file:   rosbridge-library_0.7.13.bb
    new file:   rosbridge-server_0.7.13.bb
    new file:   rosbridge-suite.inc
    new file:   rosbridge-suite_0.7.13.bb


wintermuteger:
Just noticed that though everything builds fine a dependency is missing (rosauth seems to be skipped already in the CMakeFiles.txt of rosbridge). Will try to fix it and - if succesfull - create a new pull request.

wintermuteger:
I fixed the missing ros dependencies by now. There is still a python module (bson) missing. Not sure where to get this one from, yet.

bulwahn:
I quickly looked for an existing recipe for the python bson package on layers.openembedded.org [1]. Unfortunately, it seems that there is no recipe for that package yet. So you need provide this recipe yourself. If you need help to write this recipe, I am glad to assist.

[1] http://layers.openembedded.org/layerindex/branch/master/recipes/?q=bson

wintermuteger:
I created the required recipes for the missing modules (in total it was bson, pytz and six). Unfortunately they all install with setuptools and not catkin. Therefore they are deployed to /usr/lib/python(...). 

Any idea why those modules are available when installing ROS to e.g. an Ubuntu machine? Are they part of a catkin built package (with different build-options then what is used for meta-ros) or is it because ROS modules can access all python-paths and not only the ones in /opt/ros(...) and they are part of a standard Ubuntu release?

bulwahn:
The python modules are available to ROS packages because they are part of the standard Ubuntu release. The Ubuntu desktop installation are includes many packages, much more than a minimal (Yocto Project) Poky installation. ROS packages search for python modules in /usr/lib/python and in /opt/ros/lib/python, so the modules are found if they are installed in one of these two locations.
The recipes for the python modules should be simple to set up, as there is already a setuptools class upon you can base the recipes.

wintermuteger:
Ok, I added the recipes for the missing python modules to the recipes-support section. I -c clean'ed everyhting and re-bitbake'd rosbridge-suite, which worked without errors. Default install is /usr/lib/python, but according to your explanations this should work just fine (on my test platform I manually moved them to /opt/ros/indigo/lib/python).

wintermuteger:
Ok, I think I now got nearly everything right. Just don't know how to merge multiple commits to a single one...

bulwahn:
You can merge commits with 'git rebase -i' you might want to create a temporary branch first, so that you can see how you merge the commits and still have a backup (on the original branch) while you modify the commits.

PR #370, opened 2016-01-12
Recipe updates in indigo distribution in January 2016

bulwahn:


bulwahn:
The first test failure is unrelated to the commit, but rather some git pull problem.

PR #371, opened 2016-01-13
Recipes for libccd, FCL, OMPL and OpenNI2

jvgomez:
This PR contains the recipes for OpenNI2 and OMPL Additionally, it includes recipes for libccd and FCL which might be useful when using OMPL.

PR #372, opened 2016-01-13
Redo of PR #371: Recipes for libccd, FCL, OMPL

jvgomez:
Cleaner version of PR #371 
- OpenNI2 recipe has been removed as currently assumes ARM architecture.
- FCL LICENSE file is required as it includes some notes about the copyright (it is the [same file](https://github.com/flexible-collision-library/fcl/blob/master/LICENSE) as in the original repo


bulwahn:
From my side, this looks good. @KristofRobot do you have any further points?

PR #376, opened 2016-01-18
Catkin fix

dominiquehunziker:
Add support to build catkin packages on the target platform.

The main issue I identified with cross compiled catkin packages is that various config files are generated containing the absolute paths of the host/build system. I added a pre-process step before the packaging to remove these path components. However, this will only work as long as only the standard catkin files are generated (other generated configuration files might be missed).

Additionally, I added a patch for the cmake module FindBoost which is the diff to Ubuntu's FindBoost version and I added some missing dependencies for catkin-dev.


bulwahn:
See my inline comments.

bulwahn:
I understand the issue now, and I will run some `bitbake core-image-ros-world` on the different architectures to make sure everything still works. You can consider providing these patches also on the openembedded-core mailing list, so that the issue is solved for a larger user group.

sephalon:
Could this be extended to enable cross compiling from an ADT? I tried it, but lots of stuff like all the Python libraries are missing in the x86 sysroot, and I'm also having trouble telling catkin where to find non-ROS libraries like Boost. I'd be glad for any hints I you have no time to work on this.

@dominiquehunziker Sorry for pushing this again, but do you have any advice here?

dominiquehunziker:
@bulwahn The patch for the Python dependency has been added [here](http://git.openembedded.org/openembedded-core/commit/?h=master-next&id=53376f45adeb3bf608714acc849af3c0adb9dfa3).

The FindBoost.cmake patch is obsolete with the HEAD of openembedded-core, because they are no longer using version 2.8.12 and in newer cmake versions this has been fixed.

@sephalon I'm not clear on what you want to do and therefore what the problem is.

sephalon:
@dominiquehunziker Sorry, I'll elaborate this a little bit more: Using the `populate_sdk` task, one can build an [Application Development Toolkit (ADT)](http://www.yoctoproject.org/docs/1.7/adt-manual/adt-manual.html#optionally-building-a-toolchain-installer) for an image. You can install this on your development machine to cross-compile stuff for the target machine. Basically an ADT contains two sysroots, one matching the root filesystem of your target (including all libraries and headers) and another one with the cross toolchain (compiler, linker, debugger and so on) compiled for the (most likely x86) development machine architecture. This way you could build catkin packages on-the-fly and even skip deployment by just mounting the build folder from the target using SSHFS or so.

So far I wrote a [toolchain.cmake](https://cmake.org/cmake/help/v3.0/manual/cmake-toolchains.7.html) file pointing to the cross compiler. Most importantly this has a `CMAKE_SYSROOT` variable set telling CMake to only look for stuff in the x86/development machine sysroot. When I run catkin (locally installed on my machine, but with `-DCMAKE_TOOLCHAIN_PATH` set) the compiler is detected nicely, but lots of Python dependencies like EmPy and the nosetests binary are missing in the x86 sysroot. After I just copied those from the arm into the x86 sysroot, I got a little bit further until getting errors for missing Boost libraries; and that's when I stopped, not sure what other problems are to come. Well, this post got a bit long; basically I was just hoping that you'd have some idea what has to be changed in order for Yocto to add all missing requirements to the development machine sysroot.

dominiquehunziker:
@sephalon Ok, now I get it. Didn't realize that this even existed. We do basically the same thing, but use a standalone shell script for fast cross-compiling which piggybacks on the two sysroots of bitbake. Let me check whether I can build an ADT from this.

bulwahn:
I tested this pull request with the current openembedded-core repository version, and the cmake bbappend additions lead to the following failures, as cmake is updated to 3.4.2 by now:

```
ERROR: No recipes available for:
  /home/lukas/dev/github.com/bmwcarit/meta-ros/recipes-devtools/cmake/cmake-native_2.8.12.%.bbappend
  /home/lukas/dev/github.com/bmwcarit/meta-ros/recipes-devtools/cmake/cmake_2.8.12.%.bbappend
```

I would suggest to move the needed changes for cmake 2.8.12 into a directory backports or backport-patches, and then add a section in the README.md to explain what changes must be done to apply the changes on the older versions of poky. We can certainly integrate the first patch and the other parts of the second patch (besides the cmake bbappend and related changes).

graugans:
@sephalon & @dominiquehunziker Are there any progress on the `populate_sdk` stuff? We are thinking about a ROS instance running on a embedded target and want to provide the customers a cross-toolchain which allows the customer to build his own ros packages.

dominiquehunziker:
I have something working. First, however, I have clean it up a bit and test it more thoroughly, but I don't have the time to do it just now.

graugans:
I'll start a new issue so we do not need to hijack this PR

dominiquehunziker:
This PR has been superseded by #386

PR #383, opened 2016-05-03
libccd: update to new GitHub source

jvgomez:


bulwahn:
Looks good. Thanks for your maintenance work.

PR #385, opened 2016-05-26
Drop obsolete call to sysroot_stage_libdir

simium:
It replaces the sysroot_stage_libdir() call as it is no longer available (it was just a wrapper for sysroot_stage_dir() anyway).

Please see: https://github.com/openembedded/openembedded-core/commit/80e7e7f78d957b8159bede2a5cd5614d8d73039c

### DIFF
--- a/classes/ros.bbclass
+++ b/classes/ros.bbclass
@@ -71,7 +71,7 @@ ros_sysroot_preprocess () {
         sysroot_stage_dir ${D}${ros_sysconfdir} ${SYSROOT_DESTDIR}${ros_sysconfdir}
     fi
     if [ -d ${D}${ros_libdir} ]; then
-        sysroot_stage_libdir ${D}${ros_libdir} ${SYSROOT_DESTDIR}${ros_libdir}
+        sysroot_stage_dir ${D}${ros_libdir} ${SYSROOT_DESTDIR}${ros_libdir}
     fi
     sysroot_stage_dir ${D}${ros_datadir} ${SYSROOT_DESTDIR}${ros_datadir}
     if [ -d ${D}${ros_stacksdir} ]; then


bulwahn:
Can you please refer to the meta-oe changeset where the sysroot_stage_libdir() is removed in the commit message? Can you also tell if this commit here is backwards-compatible, e.g., does it work with meta-oe versions before that changeset? Otherwise a note in the README might be useful.

simium:
@bulwahn this commit should be backwards-compatible, as sysroot_stage_libdir has always been a wrapper for sysroot_stage_dir as far as 2012 (did not bother to check older code since this file started in 2013)

bulwahn:
Thanks for your work. Can you add the reference to the meta-oe changeset _in the commit message_. I do not want to be overly picky, but usually when things break later, bisecting shows you the commit and commit message, but one does not look into the discussion in github (immediately). Hence, the important reference should be in the commit message to quickly understand the cross-dependency to openembedded-core.

simium:
@bulwahn I totally understand, don't worry. I had never edited a commit message before, I didn't know you could do that :) I think now I did it properly. Please advice otherwise. Thanks for your help.

bulwahn:
With your actions, you rebased your commit on some intermediate changeset. You probably need to rebase your branch on the current master of the bmwcarit branch. It is probably also better to put the long hyperlink simply at the end of the commit message, e.g., line 3 or 4, instead of making it part of the first line. I hope you can succeed doing these actions with git (after the basic git operations, advanced git commands have a rather steep learning curve); if you do not get it working and you get frustrated, I can do the needed git magic at some point next week.

simium:
Ok, I think I managed to do some of that magic myself and I hope it works for you now. This is my very first pull request ever and I wasn't really aware of the whole process. Thanks!

bulwahn:
@simium Thanks. This final commit looks GREAT now. 👍
@KristofRobot Any further comments? Please review and merge if it looks good for you. Thanks.

Issue #380, opened 2016-03-10
Workflow for cross-development catkin packages with meta-ros


graugans:
This issue is a follow up of the discussion with @sephalon and @dominiquehunziker on pull request #376 

I have the following scenario in mind. We develop a embedded embedded system and provide the _customers_ a toolchain for this target. The toolchain is generated by the `populate_sdk` task. This task generates a SDK suited for the image deployed on the target. It would be really cool when the user could do the following and compile packages for the target in this case the [O3D3xx ToF Camera ROS Package](https://github.com/lovepark/o3d3xx-ros) from @tpanzarella

```
$ source /opt/poky/1.8.1/environment-setup-armv7ahf-vfp-neon-poky-linux-gnueabi
$ mkdir o3d3xx-catkin
$ cd o3d3xx-catkin
$ mkdir src
$ cd src
$ catkin_init_workspace
$ ln -s ~/dev/o3d3xx-ros o3d3xx
$ catkin_make
```

In my naive way I tried to build a `nativesdk` version of catkin. To be able to build the nativesdk packages  the following patch was needed. This patch is against my fork for integration of the [libo3d3xx](https://github.com/graugans/meta-ros/tree/fido) into meta-ros

``` patch
 diff --git a/recipes-devtools/python/python-catkin-pkg_1.2.6.bb b/recipes-devtools/python/python-catkin-pkg_0.2.6.bb
index fe54e25..259c388 100644
--- a/recipes-devtools/python/python-catkin-pkg_0.2.6.bb
+++ b/recipes-devtools/python/python-catkin-pkg_0.2.6.bb
@@ -12,4 +12,4 @@ S = "${WORKDIR}/${SRCNAME}-${PV}"

 inherit distutils

-BBCLASSEXTEND = "native"
+BBCLASSEXTEND = "native nativesdk"
diff --git a/recipes-devtools/python/python-empy_3.3.2.bb b/recipes-devtools/python/python-empy_3.3.2.bb
index bfe11e4..b6a8a44 100644
--- a/recipes-devtools/python/python-empy_3.3.2.bb
+++ b/recipes-devtools/python/python-empy_3.3.2.bb
@@ -12,4 +12,4 @@ S = "${WORKDIR}/${SRCNAME}-${PV}"

 inherit distutils

-BBCLASSEXTEND = "native"
+BBCLASSEXTEND = "native nativesdk"
[...]

Issue #331, opened 2015-06-19
Buiding ROS package - catkin_make doesn't work


amoghskulkarni:
I wrote a new recipe for bare-bone ros containing ros-comm package and its dependencies. The recipe is as follows -

``` bash
DESCRIPTION = "A small image with bare-bone ROS."

IMAGE_INSTALL = "packagegroup-core-boot ${ROOTFS_PKGMANAGE_BOOTSTRAP} ${CORE_IMAGE_EXTRA_INSTALL}"

IMAGE_LINGUAS = " "

LICENSE = "MIT"

inherit core-image

IMAGE_ROOTFS_SIZE = "8192"

# remove not needed ipkg informations
ROOTFS_POSTPROCESS_COMMAND += "remove_packaging_data_files"

IMAGE_INSTALL += "packagegroup-ros-comm packagegroup-core-buildessential openssh opkg"

```

I can `bitbake` the recipe without getting any error, and run the image on the target architecture (qemu in this case). `roscore` runs fine (although for that to work, I have to tweak `/etc/hosts` and `/opt/ros/indigo/share/roslaunch/resources/roscore.xml`).

Although, `catkin_init_workspace` and `catkin_create_pkg` work; `catkin_make` fails with following output -

``` bash
root@qemuarm:~/workspace# ls
src
root@qemuarm:~/workspace# cd src/
root@qemuarm:~/workspace/src# ls
CMakeLists.txt  hello_world
root@qemuarm:~/workspace/src# cd hello_world/
root@qemuarm:~/workspace/src/hello_world# ls
CMakeLists.txt  package.xml
root@qemuarm:~/workspace/src/hello_world# cd ../..
root@qemuarm:~/workspace# catkin_make
Base path: /home/root/workspace
Source space: /home/root/workspace/src
Build space: /home/root/workspace/build
Devel space: /home/root/workspace/devel
Install space: /home/root/workspace/install
####
#### Running command: "cmake /home/root/workspace/src -DCATKIN_DEVEL_PREFIX=/home/root/workspace/devel -DCMAKE_INSTALL_PREFIX=/home/root/workspace/install -G Unix Makefiles" in "/home/root/workspace/build"
####
-- The C compiler identification is GNU 4.9.2
-- The CXX compiler identification is GNU 4.9.2
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
[...]

PR #386, opened 2016-05-27
Catkin fix v2

dominiquehunziker:
Supersedes PR: #376
Related issues: #331 #380

See [this](https://github.com/bmwcarit/meta-ros/issues/380#issuecomment-198489593) for a rough outline of the functionalities added (the issue relating to the missing libraries has been addressed).


bulwahn:
Can you comment on the few TODO notes left in the commit? The travis CI reports that two patches do not have a Upstream-Status; do you have plans to adjust the upstream ROS packages to be cross-compilable, or are you already in discussion with the ROS maintainers?

dominiquehunziker:
The TODO is not critical. It is a nice to have feature. Currently, it's also necessary to either create a setup script or add the relevant environment configuration as stated in the README.md. So, there is no loss of functionality; I just added it as a reminder.

I haven't made a decision yet whether I should try to push the two patches upstream. It is not a very clean solution to start with, i.e., initially I had a different approach in mind, but it would require bigger changes to catkin.
~~Also the current approach relies on the fact that gcc will prepend the configured sysroot automatically if the search path for header files are specified like this: `-I=/usr/include`. So, if a compiler different from gcc is used I'm not sure it will work. This might be a problem in general for this PR, not just for the suitability for an upstream PR.~~ Edit: After writing this answer I realized that I had missed an obvious fix for the search path for header files issue. Now the patch is also cleaner. Nevertheless, there is still an issue with the generated setup files as well as all the files templated on `@PYTHON_EXECUTABLE@`

bulwahn:
From reviewing the changes, there are no further concerns from my side. The travis script still reports the two failures: Upstream-Status lines in patches are a good practice in the overall Yocto project, which is relatively easy to live up to and maintain.
On Monday or Tuesday, I will have time to actually run and build some test images with this PR and check if there are some issues on my local setup.

jvgomez:
Just for the record: I was able to compile an image and also create a catkin package on the target machine. I haven't tested yet the cross-compilation.

dominiquehunziker:
Note: There is an issue with the dependencies of the catkin package. The package should only be necessary if one wants to compile code on the target machine (otherwise catkin-runtime should be suffienent). It should, therefore, depend on the *-dev of gcc, etc. to install the header files, etc. However, we had this QA error when doing that -- I'm not sure how this should be handled.

bulwahn:
@dominiquehunziker I hope that you can and will provide a new pull request with this feature. It is a good idea to get the feature running; unfortunately, I just did not have the spare time to test the pull request yet. Next time, I hope to find the time to do some minimal testing and merge the pull request.

dominiquehunziker:
Yes, didn't realize that deleting the branch closed the PR. I'm working on a new version, but I'm also a bit short on time.

In the meanwhile I'm trying to push the necessary changes to catkin itself; though the changes will only affect kinetic-devel and probably won't be available in indigo.

dominiquehunziker:
@bulwahn I don't know whether I should create a new PR for this, but [here](https://github.com/rapyuta/meta-ros) is the newest version of the catkin fix. It would be great if you could add your opinion on the [catkin PR](https://github.com/ros/catkin/pull/811), which is the same patch as applied in the meta-ros fork.

Issue #384, opened 2016-05-14
Task for catkin_runtime throws error


t0mmer:
Hello,

I am trying to do build an core-image-ros-roscore image according to "Usage" found at https://github.com/bmwcarit/meta-ros for qemux86.
I am using a common download and sstate-cache folder on my hard disk which are also used for other platforms.
With fresh clones of all needed repositories I get following error:

```
DEBUG: Executing python function sysroot_cleansstate
DEBUG: Python function sysroot_cleansstate finished
DEBUG: Executing shell function do_configure
NOTE: cmake.bbclass no longer uses OECMAKE_BUILDPATH.  The default behaviour is now out-of-tree builds with B=WORKDIR/build.
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/g++
-- Check for working CXX compiler: /usr/bin/g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Using CATKIN_DEVEL_PREFIX: /home/btr8fe/workspace/sandbox/yocto/ros_hydro/devel/build/qemux86/tmp-glibc/work/x86_64-linux/catkin-runtime-native/0.6.16-r0/build/devel
-- Using CMAKE_PREFIX_PATH: /home/btr8fe/workspace/sandbox/yocto/ros_hydro/devel/build/qemux86/tmp-glibc/sysroots/x86_64-linux/opt/ros/indigo
-- This workspace overlays: /home/btr8fe/workspace/sandbox/yocto/ros_hydro/devel/build/qemux86/tmp-glibc/sysroots/x86_64-linux/opt/ros/indigo
-- Found PythonInterp: /home/btr8fe/workspace/sandbox/yocto/ros_hydro/devel/build/qemux86/tmp-glibc/sysroots/x86_64-linux/usr/bin/python-native/python (found version "2.7.11") 
[...]

PR #387, opened 2016-06-01
catkin: patch _setup_util.py.in to address #384

bulwahn:
In _setup_util.py.in, we use python provided by environment instead of
the generated one, as the generated python path can be arbitrarily long
(depending on the location of the build directory), and hence reaches
the shebang line limit, causing the script to fail at start-up.

The catkin developers changed the cmake-generated template _setup_util.py
to use @PYTHON_EXECUTABLE@ instead of /usr/bin/env python in changeset
bf12b40c2 [1]. We revert this change here to address the issue #384 [2].

[1] https://github.com/ros/catkin/commit/bf12b40c2a93b11c3666619a474d35bcfea20837
[2] https://github.com/bmwcarit/meta-ros/issues/384

Signed-off-by: Lukas Bulwahn lukas.bulwahn@oss.bmw-carit.de

### DIFF
--- a/recipes-ros/catkin/catkin.inc
+++ b/recipes-ros/catkin/catkin.inc
@@ -9,8 +9,11 @@ SRC_URI = "https://github.com/ros/${ROS_SPN}/archive/${PV}.tar.gz;downloadfilena
 SRC_URI[md5sum] = "66f89dd5f7dbeb1662ff32890acabd70"
 SRC_URI[sha256sum] = "2cb7efdaecc238a156929861645c5cfd96977a8606fb7765e131304072d1c82c"
 
-SRC_URI += "file://0001-CATKIN_WORKSPACES-Don-t-require-.catkin-file.patch\
-            file://0002-PKG_CONFIG_LIB_PATHS-use-cmake-root-path.patch"
+SRC_URI += "\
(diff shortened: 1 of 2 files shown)


bulwahn:
@t0mmer Please check if this patch resolves the reported issue.

bulwahn:
@dominiquehunziker Please check if this change contradicts the work you did for proper SDK support.

bulwahn:
@KristofRobot Please merge once the reviewers report that this pull request resolves the issue and is reasonable.

dominiquehunziker:
For the SDK support there is an issue that @PYTHON_EXECUTABLE@ is only valid in the "native" machine; once the SDK is installed in another machine the path most likely is no longer valid. So, this is also a fix for this particular issue. => no contradiction
However, @PYTHON_EXECUTABLE@ is used also in other parts of the auto-generated catkin files, which would make this PR only a partial fix to add support for the setup scripts to the SDK.

bulwahn:
@dominiquehunziker This patch actually only tries to solve the issue #384; we should make sure that we can apply this patch and afterwards still your patches for the SDK.

bulwahn:
After the reporter of #384 did not respond the last week and it resolves the reported issue on my local machine, I consider it a working patch and will merge it.

PR #394, opened 2016-06-08
Some necessary changes to build meta-ros with the current master 

bulwahn:


bulwahn:
Only minor changes; hence, I am merging them without further review.

mdangler:
May want to add tcl to the depends in python-imaging_1.1.7.bb. While doing bitbake the bb file had errors because it could not find the package. And thank you very much since I have been blasting OE since python-imaging was not working with ROS. Cheers, Michael

mdangler:
You may want to ignore my comment. My build is on yocto/poky and it has issues.

PR #403, opened 2016-06-23
adjust some LIC_FILES_CHKSUM retroactively to version update

bulwahn:
After the version updates of recipes, I re-ran a successful build of all packages (bitbake core-image-ros-world) without a clean build. Due to some strange reason, the license check of the updated packages was skipped on the non-clean build. Some days later, i re-ran on a clean build and noticed that license files were updated. These commit address this issue.


bulwahn:

PR #405, opened 2016-07-09
urdfdom-headers: Allow empty package to be created

JochiPochi:
Short:
Solves bitbake dependency error when IMAGE_FEATURES += " dev-pkgs " is set

Long:
urdfdom-headers are only headers so bitbake packages them all into
urdfdom-headers-dev. Since there are no files to be packaged into
urdfdom-headers then it doesn't create the package. urdfdom-headers is a dependency
of urdfdom-headers-dev by default so bitbake errors when building an image
with dev-pkgs enabled.

### DIFF
--- a/recipes-ros/urdfdom-headers/urdfdom-headers_0.3.0.bb
+++ b/recipes-ros/urdfdom-headers/urdfdom-headers_0.3.0.bb
@@ -11,3 +11,5 @@ SRC_URI[sha256sum] = "f059350cd85fc3b8394ed0c3bd0d4b8d14fa9c8edd09c3ee01881b4711
 S = "${WORKDIR}/${ROS_SP}"
 
 inherit cmake ros faulty-solibs
+
+ALLOW_EMPTY_${PN} = "1"


bulwahn:
The content of the commit looks okay. :+1: Here just some stylistic remarks on the commit message: Please check the openembedded style guide (http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines); we try to follow that as far as possible. So in your case, watch out for the mandatory empty lines and line length (needed for good presentation) and always write fluent, readable English sentences. Also a reference/URL to the documentation for this bitbake feature, ALLOW_EMPTY, could be added to the commit message. Can someone of the watchers confirm that this commit does not break the usual build process (as I currently do not have access to any build machine)?

JochiPochi:
Great, thanks for the feedback! I'll edit the commit message right away :)

JochiPochi:
I changed the commit message. Hope it complies this time :)

bulwahn:
Perfect Commit! :100:

PR #410, opened 2016-09-14
Updates for recipes in recipes-devtools

bulwahn:
While ROS packages have been updated frequently, the recipes in the recipes-devtools directory had only little attention in the past. With this pull request, we provide all recipes with the currently latest release versions.


bulwahn:

PR #414, opened 2016-09-23
diagnostics: updating to 1.8.10 and adjusting dependencies

bulwahn:
In 2014, the update of the diagnostics recipes from 1.8.3 to 1.8.4 failed due to reasons that seem not to relevant anymore in the current version 1.8.10.
As the diagnostic-aggregator recipe failed with gcc6 (cf. https://github.com/bmwcarit/meta-ros/issues/392), I revisited the diagnostics recipes and updated them. The current diagnostics recipe in version 1.8.10 do not fail with gcc6, and is one step forward to build meta-ros with gcc6.


bulwahn:

bulwahn:
Thanks for catching the errors. I must have been more tired on Friday than I thought. It looks much better now.

PR #426, opened 2016-10-26
README.md: point out OpenEmbedded category on discourse.ros.org

bulwahn:


bulwahn:
@KristofRobot I will merge this commit without your review; simply because it is just a small textual edit. If you have nevertheless any remarks on the change, just add a comment here in the pull request, even though it has been merged already.

PR #428, opened 2016-10-28
python-rosdistro: add missing lib rosep2 depends on

rojkov:
roswtf reports uninitialized /etc/ros structure and recommends
to fix the issue by running `rosdep init`. But rosdep throws an
exception when trying to import the missing rosdistro library.

The library is supposed to be dropped in kinetic, but it's still
needed in indigo which the layer is based on at the moment.
So, the patch adds the library to the layer.

Signed-off-by: Dmitry Rozhkov dmitry.rozhkov@linux.intel.com

### DIFF
--- a/recipes-devtools/python/python-rosdep_0.11.5.bb
+++ b/recipes-devtools/python/python-rosdep_0.11.5.bb
@@ -12,7 +12,7 @@ SRC_URI[sha256sum] = "46e072074270a1ea25411829814f4bdc1f34b358c950e9b8a7b75b4a40
 
 S = "${WORKDIR}/${SRCNAME}-${PV}"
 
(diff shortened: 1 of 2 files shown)


bulwahn:
@rojkov Dear Dmitry, thanks for your pull request. The commit looks good. I just noticed the typo in the commit title `rosep2 -> rosdep2`, which should be easy to fix with a quick `git commit --amend`. Also, as part of this pull request you probably want to have a commit where you would like to add python-rosdistro to the rdepends of rosdistro. Once you addressed the points in any way you'd like, I am happy to merge this pull request.

rojkov:
@bulwahn Good point. I've updated the commit message and added the dependency.

bulwahn:
💯

PR #429, opened 2016-10-28
Update README.md to reflect roswtf's needs.

rojkov:


bulwahn:
@rojkov Dear Dmitry, thanks for the improvements. The first commit 18a9288 looks good. For the second commit 76d5bfa, I am wondering why you selected to suggest fedora. I would have simply proposed ubuntu which is the default distribution? But I think we should also add a further remark, that openembedded/yocto is not supported and that you must interpret the output of roswtf by manually translating the ubuntu package name to the recipe/package name that was chosen in the meta-ros layer.

rojkov:
Changed the distro name to ubuntu and added the remark about manual translation of ubuntu package names to meta-ros's recipes.

bulwahn:
Looks good. 👍

PR #444, opened 2017-01-09
librealsense: Set build type to ROS explicitly

rojkov:
The fix for build type detection in librealsense's CMakeLists.txt
turned out to be non-functional in the official ROS build farm as
explained in https://github.com/IntelRealSense/librealsense/pull/404

Since this recipe is specific to meta-ros set the build type
explicitly to ROS to guaranty the catkin-related configs for
librealsense get created and installed properly.

Signed-off-by: Dmitry Rozhkov <dmitry.rozhkov@linux.intel.com>


bulwahn:
This pull request looks good to me. I will run just a quick build of the affected recipes and then merge it.

bulwahn:
Okay, I just tested it on the current master of oe-core and meta-oe and it works. Thanks.

Issue #291, opened 2014-08-19
rosconsole: librostime.so: error adding symbols: File in wrong format


KristofRobot:
Error with `bitbake rosconsole`:

```
kristof@test:~/OE/build$ bitbake rosconsole
NOTE: Started PRServer with DBfile: /home/kristof/OE/build/cache/prserv.sqlite3, IP: 127.0.0.1, PORT: 50679, PID: 32504
Parsing recipes: 100% |#########################################| Time: 00:00:55
Parsing of 1653 .bb files complete (0 cached, 1653 parsed). 2109 targets, 70 skipped, 0 masked, 0 errors.
NOTE: Resolving any missing task queue dependencies

Build Configuration:
BB_VERSION        = "1.23.1"
BUILD_SYS         = "x86_64-linux"
NATIVELSBSTRING   = "Ubuntu-13.10"
TARGET_SYS        = "i586-oe-linux"
MACHINE           = "qemux86"
DISTRO            = "nodistro"
DISTRO_VERSION    = "nodistro.0"
TUNE_FEATURES     = "m32 i586"
TARGET_FPU        = ""
meta              = "master:d0be4b37743492fc9c178fd6f9ef73a5eb2fd9c3"
meta-oe           
meta-python       = "master:5f532a928c7df863a3f0d32db3918c8f7d227e2b"
meta-ros          = "meta-python:404893ef6761cfeccef87ee7982bae4ca3aa6595"

NOTE: Preparing runqueue
NOTE: Executing SetScene Tasks
NOTE: Executing RunQueue Tasks
ERROR: Function failed: do_compile (log file is located at /home/kristof/OE/build/tmp-eglibc/work/i586-oe-linux/rosconsole/1.10.10-r0/temp/log.do_compile.32602)
ERROR: Logfile of failure stored in: /home/kristof/OE/build/tmp-eglibc/work/i586-oe-linux/rosconsole/1.10.10-r0/temp/log.do_compile.32602
Log data follows:
| DEBUG: SITE files ['endian-little', 'bit-32', 'ix86-common', 'common-linux', 'common-glibc', 'i586-linux', 'common']
| DEBUG: Executing shell function do_compile
| NOTE: make -j 2
[...]

PR #446, opened 2017-01-10
revert commit beb4677 to fix the issue #291

rojkov:
The patch added in the reverted commit was meant to fix the issue https://github.com/bmwcarit/meta-ros/issues/291

In fact it enables CMake to look for libraries in a host system which
leads to errors when a path is tested for a library presence with the
help of CMake's find_library() command: e.g. a non-existing host
directory is tested for library presence and find_library() returns
successfully because the library exists in bitbake's sysroot; then the
directory is used by the linker, but the library doesn't exist in the
directory -> failure. In worse cases the host directory may actually exist
and contain the library, but of wrong architecture, format or
incompatible ABI making finding the root cause a difficult task.


bulwahn:
Without this patch applied, `bitbake industrial-robot-client` fails at do_compile with:
```
| ERROR: oe_runmake failed
| [ 52%] Linking CXX shared library devel/lib/libindustrial_robot_client_bswap.so
| /home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/x86_64-linux/usr/bin/cmake -E cmake_link_script CMakeFiles/industrial_robot_client_bswap.dir/link.txt --verbose=1
| /home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/x86_64-linux/usr/bin/i586-oe-linux/i586-oe-linux-g++  -fPIC  -m32 -march=i586  --sysroot=/home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/qemux86  -O2 -pipe -g -feliminate-unused-debug-types -fdebug-prefix-map=/home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/work/i586-oe-linux/industrial-robot-client/0.5.0-r0=/usr/src/debug/industrial-robot-client/0.5.0-r0 -fdebug-prefix-map=/home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/x86_64-linux= -fdebug-prefix-map=/home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/qemux86=  -fvisibility-inlines-hidden  -m32 -march=i586  --sysroot=/home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/qemux86 -Wl,-O1 -Wl,--hash-style=gnu -Wl,--as-needed -shared -Wl,-soname,libindustrial_robot_client_bswap.so -o devel/lib/libindustrial_robot_client_bswap.so CMakeFiles/industrial_robot_client_bswap.dir/src/joint_relay_handler.cpp.o CMakeFiles/industrial_robot_client_bswap.dir/src/robot_status_relay_handler.cpp.o CMakeFiles/industrial_robot_client_bswap.dir/src/joint_trajectory_downloader.cpp.o CMakeFiles/industrial_robot_client_bswap.dir/src/joint_trajectory_streamer.cpp.o CMakeFiles/industrial_robot_client_bswap.dir/src/joint_trajectory_interface.cpp.o CMakeFiles/industrial_robot_client_bswap.dir/src/robot_state_interface.cpp.o CMakeFiles/industrial_robot_client_bswap.dir/src/utils.cpp.o  -L/opt/ros/indigo/lib -Wl,-rpath,/opt/ros/indigo/lib: -lsimple_message_bswap
| /home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/x86_64-linux/usr/libexec/i586-oe-linux/gcc/i586-oe-linux/6.3.0/ld: cannot find -lsimple_message_bswap
| collect2: error: ld returned 1 exit status
| CMakeFiles/industrial_robot_client_bswap.dir/build.make:253: recipe for target 'devel/lib/libindustrial_robot_client_bswap.so' failed
| make[2]: *** [devel/lib/libindustrial_robot_client_bswap.so] Error 1
| make[2]: Leaving directory '/home/lukas/dev/openembedded.org/openembedded-core/build/tmp-glibc/work/i586-oe-linux/industrial-robot-client/0.5.0-r0/build'
| CMakeFiles/Makefile2:1220: recipe for target 'CMakeFiles/industrial_robot_client_bswap.dir/all' failed
| make[1]: *** [CMakeFiles/industrial_robot_client_bswap.dir/all] Error 2
```
With the patch applied, the issue disappears. Unfortunately, in my build configuration, do_compile then fails with some further type errors in boost libraries, probably due to stricter type enforcement with gcc-6.

bulwahn:
I believe that the problem is more a problem in the CMakeLists.txt of industrial_robot_client rather than issue #291. For example, the line
```
target_link_libraries(industrial_robot_client_bswap simple_message_bswap)
```
should be changed not to use the string `simple_message_bswap` but to refer to this library with a variable, probably `${simple_message_bswap_LIBRARIES}`.

bulwahn:
Around the history of commit beb4677, there should be some commits in the meta-ros repository that patch CMakeLists.txt in some ROS packages that did not use variables but hard-coded strings. I remember that we uncovered a couple of those incidences a few years ago.

rojkov:
I'm using gcc5 at the moment and it's more relaxed about the types indeed.

There's the issue https://github.com/ros-industrial/industrial_core/issues/46 filed for industrial_robot_client. Actually it stems from the package `simple_message` which provides multiple libraries built from the same source files (and as a part of a single project), but with different build-time configs. According to the devs they have reasons not to use the usual way of linking for ROS.

Still I think that having the patch applied to catkin is dangerous regardless of the ROS-Industrial implications. It might happen that somebody less lucky than us had installed a library of the same name on host. And I'm talking not about libsimple_message.so only, but any library under catkin's control the user links to (e.g. libhelloworld.so or libtoo_generic_name.so). Then this wrong library would be used for linking with unpredictable results.

bulwahn:
I am not fully convinced that reverting the commit is the right thing to do. As we created the patch in 2014, the intension of the patch was to avoid to pick up the wrong libraries while configuring. Of course, many things have changed in meta-ros and the other layers since then, and your reasoning sounds valid, but I am still wondering why reverting solves the issue. In the beginning, we created the patch because of a certain problem and the patch solved that problem. Now, a very similar build problem occurs with the patch applied, and the problem is solved by removing it...
I am going to merge this pull request nevertheless. We will see if some users run into related build problems in the future.

Issue #343, opened 2015-07-25
sphinxbase: QA Issues - libpython2/libasound/libsndfile1 [build-deps]


KristofRobot:
```
WARNING: QA Issue: sphinxbase rdepends on libpython2, but it isn't a build dependency? [build-deps]
WARNING: QA Issue: sphinxbase rdepends on libasound, but it isn't a build dependency? [build-deps]
WARNING: QA Issue: sphinxbase rdepends on libsndfile1, but it isn't a build dependency? [build-deps]
```

Issue #400, opened 2016-06-20
sphinxbase: linking libsphinxad.so fails due to undefined reference in do_compile


bulwahn:
The recipe sphinxbase currently fails with the build configuration:

```
Build Configuration:
BB_VERSION        = "1.31.0"
BUILD_SYS         = "x86_64-linux"
NATIVELSBSTRING   = "Ubuntu-14.04"
TARGET_SYS        = "i586-oe-linux"
MACHINE           = "qemux86"
DISTRO            = "nodistro"
DISTRO_VERSION    = "nodistro.0"
TUNE_FEATURES     = "m32 i586"
TARGET_FPU        = ""
meta              = "master:fba321ae9a1916d64f5c49c65c60c276f3346e7a"
meta-oe           
meta-multimedia   
meta-python       = "master:9cd117a5dc5f26133a855d1cc3bf9604ae178685"
meta-ros          = "ci:c8b0688503be4f30e9ea525ab6b8cf645a2dc78f"
```

The complete log is:

```
| Making all in sphinx_adtools
| make[2]: Entering directory `/home/lukas/work/repositories/openembedded.org/openembedded-core/build/tmp-glibc/work/i586-oe-linux/sphinxbase/0.8-r0/build/src/sphinx_adtools'
| ../../i586-oe-linux-libtool  --tag=CC   --mode=link i586-oe-linux-gcc  -m32 -march=i586 --sysroot=/home/lukas/work/repositories/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/qemux86  -O2 -pipe -g -feliminate-unused-debug-types -fdebug-prefix-map=/home/lukas/work/repositories/openembedded.org/openembedded-core/build/tmp-glibc/work/i586-oe-linux/sphinxbase/0.8-r0=/usr/src/debug/sphinxbase/0.8-r0 -fdebug-prefix-map=/home/lukas/work/repositories/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/x86_64-linux= -fdebug-prefix-map=/home/lukas/work/repositories/openembedded.org/openembedded-core/build/tmp-glibc/sysroots/qemux86=   -Wl,-O1 -Wl,--hash-style=gnu -Wl,--as-needed -o sphinx_cont_adseg cont_adseg.o ../../src/libsphinxbase/libsphinxbase.la ../../src/libsphinxad/libsphinxad.la -lpthread -lm
[...]

Issue #344, opened 2015-07-25
pocketsphinx: QA Issue - libsndfile1 [build-deps]


KristofRobot:
```
WARNING: QA Issue: pocketsphinx rdepends on libsndfile1, but it isn't a build dependency? [build-deps]
```

herb-kuta-lge:
@KristofRobot < Can this issue be closed?

herb-kuta-lge:
"Silence implies consent" (see http://wiki.c2.com/?SilenceImpliesConsent).

PR #447, opened 2017-01-17
sphinxbase: add build dependencies to sphinxbase

rojkov:
sphinxbase is linked to libsndfile1, libasound from alsa-lib and
libsamplerate0, but these build dependencies are missing in the
recipe.

This update adds the missing dependencies.

Signed-off-by: Dmitry Rozhkov <dmitry.rozhkov@linux.intel.com>

### DIFF
--- a/recipes-extended/cmusphinx/sphinxbase_0.8.bb
+++ b/recipes-extended/cmusphinx/sphinxbase_0.8.bb
@@ -12,6 +12,10 @@ require cmusphinx.inc
 
 SRC_URI += "file://0001-TESTS-srcdir-remove.patch"
 
+DEPENDS = "libsndfile1 alsa-lib libsamplerate0"
+
+EXTRA_OECONF = " --without-python"
+
 do_install_append () {
     #remove egg-info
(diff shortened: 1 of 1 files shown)


bulwahn:
Thanks for looking into the issues of the sphinxbase recipe. We have been recording issues #343, #344 and #400 about sphinxbase and its dependent recipe ros-pocketsphinx, but never got to address them.

Are these three issues resolved with your commit?

Then you can add `(resolves #343, #344, #400)` to your commit message, which should resolve them in the issue tracker automagically once the pull request is merged.

bulwahn:
Accidently closed the pull request by clicking on the wrong button...

rojkov:
I've added the `resolves` tag for #343 and disabled python support to actually address it. #344 must be an overlinking issue. IIRC OE has an option preventing generation of la-files to prevent it.

I've looked at sphinxbase's `configure.in` and it seems #400 needs extra care: it links to what ever is available in the build environment. If both PulseAudio and ALSA are present it chooses PulseAudio. I think the recipe has to be reworked eventually to add conditional builddeps depending on desired features. But ultimately it can be fixed when OE starts to support isolated builds (that's work in progress).

Issue #458, opened 2017-03-07
python-nose conflicts with python3-nose 


yongli3:
During the ros-world build, there is a conflict about python-nose and python3-nose. Since opencv dpeneds on python3-nose, by tf and rostest depend on python-nose. bitbake will install both of python3-nose and python-nose, this causes the ""that file is already provided by package" error message.

I think we need to change the python-nose to python3-nose in the DEPENDS in meta-ros

bulwahn:
Thanks for reporting this issue. I am happy to review pull requests that address this issue. We should make sure that all ROS packages still work with python3, which sounds like a quite large task to me right now, but we will see once we start tackling this issue.

yongli3:
Thanks @bulwahn, please check on PR #459

yongli3:
As a workaround, we can modify the local.conf to disable python3 in opencv:
PACKAGECONFIG_remove_pn-opencv = " python3"
PACKAGECONFIG_append_pn-opencv = " python"

rojkov:
I guess I'm volunteering to check if python3 is enough for ROS because the project I'm working on [1] wants to get rid of python2 completely.

[1] https://github.com/intel/intel-iot-refkit/

bulwahn:
Well that sounds like a bigger task...

bulwahn:
@yongli3 We have now included python3 support. Could you try if the issue still occurs in your image build when you set meta-ros to use python3?

PR #459, opened 2017-03-07
Fix the python3 conflict issues

yongli3:
Since the latest opencv package depends on python3, we need to switch to python3 in meta-ros,
to fix the below bitbake error message in issue #458: “that file is already provided by package”

Signed-off-by: Yong Li <sdliyong@gmail.com>

### DIFF
--- a/recipes-ros/geometry/tf_1.11.8.bb
+++ b/recipes-ros/geometry/tf_1.11.8.bb
@@ -9,4 +9,4 @@ require geometry.inc
 
 SRC_URI += "file://0001-Fix-stdlib.h-No-such-file-or-directory-errors-in-GCC.patch;striplevel=2"
 
-RDEPENDS_${PN} = "python-numpy"
+RDEPENDS_${PN} = "python3-numpy"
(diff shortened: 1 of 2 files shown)


rojkov:
I've checked quickly the recipe for the latest opencv and it seems it can depend on either python2 or python3 depending on what's included in its `PACKAGECONFIG`.

If I'm not mistaken the proper solution would be to replace `python-nose` to `${PYTHON_PN}-nose`.

yongli3:
@rojkov I just tested the ${PYTHON_PN}, but it still uses the python-nose, is there any other changes are needed? 

Just found the catkin.bbclass uses distutils-base, it will select python 2.7. There are several files are depending on it.

bulwahn:
Just for experimenting, what happens if you replace distutils-base with distutils3-base in catkin.bbclass. How many things break in meta-ros?

yongli3:
@bulwahn After changed the distutils-base to distutils3-base in catkin.bbclass, the catkin.bbclass build will fail, almost all other packages depend on it

bulwahn:
Okay, so we really must see if it is feasible to get everything running with python3 or simply set python2 in the opencv recipe with a .bbappend file, right?

yongli3:
@bulwahn Yes. I think for the quick solution, we can modify opencv to use python2 with the .bbapend or local.conf file: 
PACKAGECONFIG_remove_pn-opencv = " python3"
PACKAGECONFIG_append_pn-opencv = " python"

bulwahn:
So, the proposed commit does not provide a solution to use either python2 or python3 throughout meta-ros. I suggest to close this pull request, and consider if we want to add some documentation in the ROS wiki or in the README.md file. The discussion can be continued in the issue #458.

PR #476, opened 2017-04-10
realsense_camera: Bump to version 1.8.0 (new stable version)

mbelluzzo:
Also, adds 'rgbd-launch' to RDEPENDS and add patch to fix linkage against system librealsense.


recipes-extended/realsense-camera/realsense-camera/0001-librealsense-Fix-link-when-the-system-library-is-use.patch
```diff
@@ -0,0 +1,30 @@
...
+index a1ec2f6..926992a 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -106,6 +106,12 @@ add_library(${PROJECT_NAME}_nodelet src/base_nodelet.cpp src/r200_nodelet.cpp sr
+ target_link_libraries(${PROJECT_NAME}_nodelet
+   ${catkin_LIBRARIES}
+ )
++if(USE_SYSTEM_LIBREALSENSE)
++  target_link_libraries(${PROJECT_NAME}_nodelet
++    ${realsense_LIBRARY}
```
rojkov, line 24: I'd prefer to have the `realsense::realsense` alias here instead of `${realsense_LIBRARY}` like the way it's done for the `get_debug_info` executable.

recipes-extended/realsense-camera/realsense-camera_1.8.0.bb
```diff
@@ -2,14 +2,13 @@ DESCRIPTION = "ROS driver for RealSense camera"
 LICENSE = "BSD"
 LIC_FILES_CHKSUM = "file://package.xml;beginline=9;endline=9;md5=4bcd5f0e140091c68f8c0bcfdd6181e0"
 
-DEPENDS = "librealsense cv-bridge camera-info-manager pcl-ros roscpp"
+DEPENDS = "librealsense cv-bridge camera-info-manager pcl-ros roscpp rgbd-launch"
```
rojkov: Probably you wanted to add `rgdb-launch` to the `RDEPENDS` field, not `DEPENDS`?

rojkov (changes requested):
Thank you for the update! Needs a couple of tweaks, otherwise looks good.

recipes-extended/realsense-camera/realsense-camera/0001-librealsense-Fix-link-when-the-system-library-is-use.patch
```diff
@@ -0,0 +1,30 @@
+From fb74e5eec2368f20c4b43cf9224519b29aee35ad Mon Sep 17 00:00:00 2001
+From: Murilo Belluzzo <murilo.belluzzo@intel.com>
+Date: Wed, 5 Apr 2017 17:06:44 -0700
+Subject: [PATCH 3/3] librealsense: Fix link when the system library is used
+
+Otherwise librealsense won't be listed as a dependency for the nodelet
+that will fail to load due to undefined symbols.
+
```
rojkov, line 8: And the patch is missing `Upstream-Status`.

recipes-extended/realsense-camera/realsense-camera_1.8.0.bb
mbelluzzo: True that! I will update this soon.

recipes-extended/realsense-camera/realsense-camera/0001-librealsense-Fix-link-when-the-system-library-is-use.patch
mbelluzzo: Is there an example on how the status should be stated on the message? It is merged upstream but is not on any release yet.

recipes-extended/realsense-camera/realsense-camera/0001-librealsense-Fix-link-when-the-system-library-is-use.patch
mbelluzzo: The problem of using "realsense::realsense" directly is that only works with master for now (there is a lot of custom Findrealsense.cmake out there).

Also is a good cmake practice to use the variables that find_package() is supposed to set instead of explicitly state the supposed values of those. Maintain consistency with other uses and is more future proof.

Another pro is that keep this patch exactly the same of what was merged upstream.

rojkov: Ok then, I see it's a bit too late to fix the patch.

But if you care about the various custom Findrealsense.cmake in the wild then a better solution would be to instruct this CMakeList.txt to ignore all these modules with
```
find_package(realsense REQUIRED NO_MODULE)
```
The librealsense recipe in all branches from fido to morty already provides proper realsenseConfig.cmake which I created exactly to encourage adoption of good cmake practices among librealsense users.

What other uses are you talking about?
The doc page for `find_package()` doesn't state that it's supposed to set anything like that. It's up to good will of a module's author to conform the convention and to set the variable. In fact the manual page [1] suggests that Xxx_LIBRARY "should not usually be used in CMakeFiles.txt". In addition to that I find imported targets (sorry, i mistakenly used the word "alias" for them) more readable than variables not set in the file where they are used. 

For more info on good practices I'd recommend reading [2] and [3].

[1] https://cmake.org/cmake/help/v3.0/manual/cmake-developer.7.html#modules
[2] https://www.slideshare.net/DanielPfeifer1/cmake-48475415
[3] https://archive.fosdem.org/2013/schedule/event/moderncmake/attachments/slides/258/export/events/attachments/moderncmake/slides/258/cmake_fosdem_2013.pdf

mbelluzzo: I care because is the only working option for most of the projects relying on librealsense stable versions up to 1.12.1.

If you take a look at your first reference, it states the above (in the plural form, to be honest, that is what I actually find on all custom FindRealsense.cmake I have to deal frequently, besides other packages also returning on this form).

[2] and [3] make use of this horrible idea without any good justification. But the fact is that it isn't widely used. Just grep all .cmake files on work dir of a a yocto build and see.

The config should return the namespace form indeed, but the user should rely on the "expected" output (I say "expected" because oficial documentation of this function is terrible like the overall cmake documentation usually is, although having improved a lot along the years) of find_package() as stated in [1] and not try to guess the value. 

And last but not least, if wasn't a common practice, why would you care to set the value of <package>_LIBRARY (going against your own references) as the last step of the cmake config that you wrote for librealsense?  :-/.

This complains makes no sense to me, it should actually be s/realsense_LIBRARY/realsense_LIBRARIES/ and a patch on your config upstream.

mbelluzzo:
branch updated

rojkov (approved):

rojkov:
I successfully built the PR rebased on top of the latest master
```
Build Configuration:
BB_VERSION        = "1.33.3"
NATIVELSBSTRING   = "fedora-25"
MACHINE           = "qemux86"
TARGET_FPU        = ""
meta              = "master:eeb17d64ecae7442a627824ccacca83959cb32e0"
meta-oe           
meta-python       
meta-multimedia   = "master:cdb428e7c49899675ee7b7a43f6cecdb5b4c2546"
meta-ros          = "mbelluzzo-nodelet-v2:c26b62cadffc08fbe59ebec184818033ce277c96"
meta-intel-realsense = "master:53e93ed3f9fad6aa96ca455172e4919b666dca3b"

```

rojkov:
I care because is the only working option for most of the projects relying
> on librealsense stable versions up to 1.12.1.
>
I see. I thought you were talking about other uses of realsense_camera to
be consistent with beyond the contexts of ROS and meta-ros. Well, sorry for
those projects they have no other choice, but to come up with their own
custom find modules as librealsense authors didn't care about their users'
convenience.

In ROS the package relies on librealsense's configs generated by catkin and
doesn't need a find module neither.


> If you take a look at your first reference, it states the above (in the
> plural form, to be honest, that is what I actually find on all custom
> FindRealsense.cmake I have to deal frequently, besides other packages also
> returning on this form).
>
> [2] and [3] make use of this horrible idea without any good justification.
> But the fact is that it isn't widely used. Just grep all .cmake files on
> work dir of a a yocto build and see.
>

Horrible? Why? How would you get relocatable and debug/release builds
without imported targets, but the "expected" variables? Remember
Xxx_LIBRARIES "should include full paths".


> The config should return the namespace form indeed, but the user should
> rely on the "expected" output (I say "expected" because oficial
> documentation of this function is terrible like the overall cmake
> documentation usually is, although having improved a lot along the years)
> of find_package() as stated in [1] and not try to guess the value.
>

Where's the guess? I know for sure librealsense's config provides an
imported target.


> And last but not least, if wasn't a common practice, why would you care to
> set the value of _LIBRARY (going against your own references) as the last
> step of the cmake config that you wrote for librealsense? :-/.
>

That was a thoughtless copy-paste from the FOSDEM presentation. Redundant
indeed.


> This complains makes no sense to me, it should actually be
> s/realsense_LIBRARY/realsense_LIBRARIES/ and a patch on your config
> upstream.
>
>
realsense_LIBRARY is not a list of strings, but another reference to the
same realsense::realsense target. Thus singular.

mbelluzzo:
I think you still are missing the most important point that when one is
writing a CMakeList.txt and using fing_package(), you don't want to care if
find package relied on a cmake config to find your dependency, if it found
via pkg-config or if it found by using custom Find*. You just want to add
your deps to the linking list and you includes. And the way to not care
about that (IOW if the return will be either "realsense::realsense or
"-llibrealsense -lOpenwhatever -lboost") is to use
<your_depedency>_LIBRARIES. It is costant, it doesnt change it, if your
find_package worked, this is expected to work with targe_link_libraries.

PR #488, opened 2017-04-25
moveit-ros-planning: call python3 for dynamic reconfigurations

rojkov:
If both python3 and python2 are available in build environment then according to PEP-394 python should reference python2. But in case of building a python3 flavor of meta-ros PYTHONPATH is set explicitly to python3 libraries. When python2 is called with PYTHONPATH pointing to python3 libs the following error triggers:

    Could not find platform independent libraries <prefix>
    Could not find platform dependent libraries <exec_prefix>
    Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]

This patch makes the dynamic reconfigurations call python3 explicitly in case a python3 flavor of  meta-ros is being built.

### DIFF
--- a/recipes-ros/moveit/moveit-ros-planning/0001-call-python3-for-dynamic-reconfigurations.patch
+++ b/recipes-ros/moveit/moveit-ros-planning/0001-call-python3-for-dynamic-reconfigurations.patch
@@ -1,0 +1,71 @@
+From 5aa9a6a148b91380210a79f5a48ed4823401e9c0 Mon Sep 17 00:00:00 2001
+From: Dmitry Rozhkov <dmitry.rozhkov@linux.intel.com>
+Date: Tue, 25 Apr 2017 10:41:56 +0300
+Subject: [PATCH] call python3 for dynamic reconfigurations
+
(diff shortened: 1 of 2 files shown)


bulwahn:
I assume it is a runtime failure that only occurs when running image, so I did not check if that is resolved with this commit, and just trust you on that.
For the usual sanity checking, I built core-image-ros-world successfully with ROS_USE_PYTHON3 = "yes" on:
```
Build Configuration:
BB_VERSION        = "1.33.4"
MACHINE           = "qemux86"
TARGET_FPU        = ""
meta              = "master:e584be78f92ee6f08f570c239698d56ac78d05f9"
meta-oe           
meta-multimedia   
meta-python       = "master:2c21aa8633e4e3538099bcc56d76ab0a594411b4"
meta-ros          = "rojkov-moveit-python3-v1:2b9f864ffab2680a303a0f278c0ad01df7aceb23"
```

rojkov:
@bulwahn Thank you for testing!
Those *.cfg files are not installed on target, but run at build time to generate other python scripts AFAIU.

PR #504, opened 2017-05-09
Fix pythonpath for systemd environment file

### LINKED ISSUE #503: PYTHONPATH in roscore systemd service file
The PYTHONPATH in the [EnvironmentFile](https://github.com/bmwcarit/meta-ros/blob/master/recipes-ros/ros-comm/roslaunch/roscore-default) of [roscore.service](https://github.com/bmwcarit/meta-ros/blob/master/recipes-ros/ros-comm/roslaunch/roscore.service#L6) should be 
`/opt/ros/kinetic/lib/python2.7/dist-packages/` instead of `/opt/ros/kinetic/lib/python2.7/site-packages/`.

mcopejans:
Closes #503

### DIFF
--- a/recipes-ros/ros-comm/roslaunch/roscore-default
+++ b/recipes-ros/ros-comm/roslaunch/roscore-default
@@ -6,4 +6,4 @@ ROS_MASTER_URI=http://localhost:11311
 CMAKE_PREFIX_PATH=/opt/ros/indigo/
 PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin:/opt/ros/indigo/bin
 LD_LIBRARY_PATH=/opt/ros/indigo/lib
-PYTHONPATH=/opt/ros/indigo/lib/python2.7/site-packages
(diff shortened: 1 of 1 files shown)


mcopejans:
@bulwahn PTAL

rojkov:
Why this change is needed?

I've just checked the path to ROS python libs in `core-image-ros-world` and it seems all the libs are inside `/opt/ros/indigo/lib/python2.7/site-packages`.

bulwahn:
@mcopejans Did you check the state on your development machine, or in the image that is created with the `bitbake` tool? I just want to make sure that there is not a misunderstanding in which this script is intended to be executed.

rojkov:
Did you test the proposed change? I believe with this patch applied `roscore` simply won't find its python libs.

AFAIK `dist-packages` is a convention specific to debian-based distros like Ubuntu. OpenEmbedded doesn't adhere this convention.

mcopejans:
That sounds like a reasonable explanation. I am using Ubuntu. Sorry for the misunderstanding. I will close the PR and linked issue.

PR #509, opened 2017-05-16
rosbridge-suite: Replace bson with pymongo, update runtime dependencies

johannesschrimpf:
Removed python-bson since bson is provided by python-pymongo in meta-python

Added runtime dependencies. Not all runtime dependencies are automatically added, which leads to
dependency problems in the package management. Added
RDEPENDS to rosbridge-library and rosbridge-server.


bulwahn:
Commit `rosbridge-server: added RDEPENDS` looks reasonable to me.
For the commit `python-bson: Removed package` and `rosbridge library: Changing dependencies`, I am wondering if python-pymongo is a much larger package than the self-contained python-bson package. Does the rosbridge-library really need the python-pymongo library? Keeping python-bson in meta-ros is not an issue, as maintenance effort is low anyway.
Other than that one point, the rest looks good to me.

johannesschrimpf:
I agree that the overhead should be considered. From my first look, it seems that the python-pymongo does not have dependencies to other packages.

I see the benefits as follows:
- Reducing complexity by following the meta-python layer
- Possible version conflicts when both python-pymongo and python-bson is installed

The following reason speaks for keeping the python-bson:
- The package.xml is dependent on python-bson directly

I see that larger distributions like ubuntu have both packages, where python-pymongo depends on python-bson. That could be a suggestion for the meta-python layer.

johannesschrimpf:
I added two commits to solve the following runtime dependencies:

Traceback (most recent call last):
  File "/opt/ros/kinetic/lib/rosbridge_server/rosbridge_websocket", line 39, in <module>
    from tornado.ioloop import IOLoop
  File "/opt/ros/kinetic/lib/python2.7/site-packages/tornado/ioloop.py", line 46, in <module>
    from tornado.log import app_log, gen_log
  File "/opt/ros/kinetic/lib/python2.7/site-packages/tornado/log.py", line 37, in <module>
    from tornado.escape import _unicode
  File "/opt/ros/kinetic/lib/python2.7/site-packages/tornado/escape.py", line 38, in <module>
    import html.entities as htmlentitydefs  # py3
ImportError: No module named html.entities

Traceback (most recent call last):
  File "/opt/ros/kinetic/lib/rosapi/rosapi_node", line 36, in <module>
    from rosapi import proxy, objectutils, params
  File "/opt/ros/kinetic/lib/python2.7/site-packages/rosapi/params.py", line 37, in <module>
    from json import loads, dumps
ImportError: No module named json

Tested with 0.7.17 in my kinetic setup but verified in source code that it also applies for 0.7.16.

rojkov:
In my opinion the maintenance burden of the separate python-bson is negligible comparing to additional build overhead, but in this particular case I'm not against this change. Though I'd expect a patch to meta-python adding a recipe for python-bson.

Still the python dependencies need to be adjusted to accommodate python3: `python-<pkgname>` -> `${PYTHON_PN}-<pkgname>`.

johannesschrimpf:
Here an example for a FileExistsError when installing both pymongo and bson. To reproduce, just include python-pymongo in your image as well as rosbridge-library:

```
ERROR: rosbridge-library-0.7.17-r0 do_prepare_recipe_sysroot: Error executing a python function in exec_python_func() autogenerated:

The stack trace of python calls that resulted in this exception/failure was:
File: 'exec_python_func() autogenerated', lineno: 2, function: <module>
     0001:
 *** 0002:extend_recipe_sysroot(d)
     0003:
File: '/data/yocto-var-pyro/poky/meta/classes/staging.bbclass', lineno: 592, function: extend_recipe_sysroot
     0588:                    dest = newmanifest[l]
     0589:                    if l.endswith("/"):
     0590:                        staging_copydir(l, targetdir, dest, seendirs)
     0591:                        continue
 *** 0592:                    staging_copyfile(l, targetdir, dest, postinsts, seendirs)
     0593:
     0594:    for f in fixme:
     0595:        if f == '':
     0596:            staging_processfixme(fixme[f], recipesysroot, recipesysroot, recipesysrootnative, d)
File: '/data/yocto-var-pyro/poky/meta/classes/staging.bbclass', lineno: 233, function: staging_copyfile
     0229:        os.symlink(linkto, dest)
     0230:        #bb.warn(c)
     0231:    else:
     0232:        try:
 *** 0233:            os.link(c, dest)
     0234:        except OSError as err:
     0235:            if err.errno == errno.EXDEV:
     0236:                bb.utils.copyfile(c, dest)
     0237:            else:
Exception: FileExistsError: [Errno 17] File exists: '/data/yocto-var-pyro/build/tmp/sysroots-components/cortexa9hf-neon/python-bson/usr/lib/python2.7/site-packages/bson/__init__.pyc' -> '/data/yocto-var-pyro/build/tmp/work/cortexa9hf-neon-poky-linux-gnueabi/rosbridge-library/0.7.17-r0/recipe-sysroot/usr/lib/python2.7/site-packages/bson/__init__.pyc'

ERROR: rosbridge-library-0.7.17-r0 do_prepare_recipe_sysroot: Function failed: extend_recipe_sysroot
ERROR: Logfile of failure stored in: /data/yocto-var-pyro/build/tmp/work/cortexa9hf-neon-poky-linux-gnueabi/rosbridge-library/0.7.17-r0/temp/log.do_prepare_recipe_sysroot.19671
ERROR: Task (/data/yocto-var-pyro/poky/meta-ros/recipes-ros/rosbridge-suite/rosbridge-library_0.7.17.bb:do_prepare_recipe_sysroot) failed with exit code '1'
```

In my opinion this is quite critical and may break setups.
I contacted the yocto developers and I am right now testing a patch from them that splits pymongo into two packages, bson being one of them. Until this is added, I suggest using pymongo to prevent incompatibility.

johannesschrimpf:
I'll check the python3 dependencies.

johannesschrimpf:
It looks like python3-twisted and python3-imaging are not available.

Also, I can't see in the rosbridge source code why python-six is needed as dependency.

bulwahn:
@johannesschrimpf I have seen that you made some changes and progress on meta-oe (without looking into details). So, what remains open to be changed in meta-ros? Or are the changes on meta-oe already sufficient and there is nothing further to do in meta-ros?

johannesschrimpf:
I haven't used rosbridge since summer, so I don't have tested it since then. However, I see that python-bson is now official part of meta-oe rocko after a feature request I made earlier this year, see here:

https://github.com/openembedded/meta-openembedded/blob/rocko/meta-python/recipes-devtools/python/python-pymongo.inc

I see that this change is not backported to pyro. It might be possible to set DEFAULT_PREFERENCE to "-1" in the meta-ros recipe to ensure compatibility with pyro while using the meta-os recipe in rocko but I'm not sure if this works since the recipe names are different, but this should somehow be addressed in meta-ros to ensure compatibility is someone uses both pymongo and bson.

Also, as mentioned before, not all dependencies can be satisfied for python3-imaging. It seems like python3-twisted has been added since I checked last, at least for rocko, but not for pyro.

So what needs to be done is the following:
 - Find a strategy to use the provided python-bson to not have conflicts and find a solution for pyro and older versions
 - Review and add RDEPENDS to meta-ros, with a strategy for the python packages that are not available for python3 in pyro and older

johannesschrimpf:
@bulwahn As I can see, the following problems are present at this time:

rocko:
 - Rocko detects that python-bson and python-pymongo (official) conflict since pymongo includes bson
 - python-bson needs to be in RDEPENDS, not DEPENDS if using the new python-bson from rocko
 - Solution for rocko: Delete python-bson, add it to RDEPENDS of rosbridge

pyro (and older):
 - python-bson not present in openembedded pyro, so it's needed in meta-ros even though it conflicts of sombody installs python-pymongo

Possible solutions:
 - 1) Use pymongo, delete bson recipe. Should work on both rocko and pyro (and older)
 - 2) Have different branches, such as most other meta layer. I suggest indigo-pyro, indigo-rocko etc. Later for kinetic then kinetic-pyro, kinetic-rocko etc.

I think ii. would be the best way, however i. would be the faster solution for now.

The runtime dependencies should be fixed anyways. rocko seems to be more sensitive to RDEPENDS.

bulwahn:
@johannesschrimpf Thank you very much for looking into this. Considering option ii, currently I did not even officially start to maintain two branches for indigo and kinetic, as the maintenance will increase for two branches. The kinetic branch is experimental with a disclaimer that I can rebase, squash commits and modify the history at any point in time.
[I will make the kinetic branch an official branch soon-ish, and won't modify the history then, but it will take a week of some clean-up activities first.]
So, currently option ii sounds like a maintenance hell to me as long as there is not a maintainer team (currently, it is only me and Dmitry and we both are only part-time available) that is reactive to the pull requests that might come.

I would like to keep the recipe DEPENDS and RDEPENDS in sync with the information in the package.xml [1] as far as possible. So, hence I would like to keep the RDEPENDS on python-bson, not on python-pymongo.

Hence, I would propose the following alternative to option i (everything else remains as in option i):
  - Keep python-bson recipe and set `DEFAULT_PREFERENCE` to -1.
Then it should work for versions before rocko, simply using the python-bson recipe, and for rocko and later, it relies on the externally provided recipe. Once, we consider not to support any version before rocko anymore, we simply delete the recipe; as a reminder, we simply add an issue to delete the python-bson recipe once we consider it appropriate.

I see two more dangling issues in the pull request:
  - Figuring out why python-six was added in the recipe as dependency (by looking at the history of this git repository and the package.xml of previous versions.), and e.g., if it was deleted at some point in the package.xml for some reason.
  - Checking the effort to make this recipe work with python3. It might just need a bit of tuning of the meta-python layer.

[1] https://github.com/RobotWebTools/rosbridge_suite/blob/develop/rosbridge_library/package.xml

johannesschrimpf:
@bulwahn 
Considering that most layers just have branches for the different yocto versions, and kinetic/indigo branches may lead to confusion, together with the edge case of system where different yocto versions may be installed/developed/tested at the same time, I would suggest looking into redesigning the layer with recipe names like in ubuntu, which means that indigo and kinetic coexist in one layer with recipe names like "ros-kinetic-catkin" and "ros-indigo-catkin".

I would imagine that could make maintenance easier, at least in some respect. Also, the change to a new version seems to be easier, as well as discontinuing old releases and adding support for new (long-term) releases. It also would be more consistent with the layer index, I guess. I don't see how to integrate meaningful search for recipes if there are different branches for kinetic and indigo.


Regarding RDEPENDS vs DEPENDS. I agree holding the dependencies as much as possible like in the package.xml. However, the first priority should be to have a working distribution of ros in yocto/openembedded. Some of the problems, I have had are:
 - Due to missing rdepends, I had to do several iterations adding recipes that were not set as rdepends. Since runtime dependencies may happen in edge cases, for embedded systems, especially in production systems, these should be addressed in the meta layer to increase stability since you may not have the build server available when finding the dependency
 - In the case of pymongo/bson there is a special case: package.xml has python-bson as both build and runtime dependency

 If this is right, this leads to the following cases

1) 
DEPENDS: python-bson
RDEPENDS: python-bson

pyro: works since python-bson is only present in meta-ros. But: If any other package depends on python-pymongo you get conflict. Risky.

rocko: First challenge: python-bson from meta-ros must be deactivated due to conflict with python-pymongo (both install bson to the same place). Second challenge: Even if python-bson is deacticated/masked in meta-ros, it is not possible to add python-bson to DEPENDS due to python-bson being part of python-pymongo and not a recipe-name


2)
DEPENDS: python-pymongo
RDEPENDS: python-bson

pyro: Conflict if python-pymongo from meta-python is used together with python-bson from meta-ros
rocko: Works if python-bson from meta-ros is masked (maybe possible using DEFALUT_PREFERENCE, but I'm not sure if possible due to recipe name/package name)


3) 
DEPENDS: python-pymongo
RDEPENDS: python-pymongo

pyro: stable if python-bson is removed from ros (but not consistent with package.xml)
rocko: stable if python-bson is removed from ros (but not consistent with package.xml)


However, the question, I would ask is whether python-bson really is a build dependency of rosbrigde. From my first look at it I cannot see why this should be and I also see that meta-ros builds fine with not using this build dependency. If it really is not a build dependency, a change request should be sent upstream. In this case we still have a conflict in pyro when other packages use python-pymongo. From my point of view, the best solution is to omit the build dependency, as done right now, and for rocko use python-bson in RDEPENDS, and for pyro python-pymongo.

Since there is only one layer I would suggest using python-pymongo for now, and at end-of-support for morty/pyro change to python-bson.

For the way further, I would suggest the following:
 1) Deciding on a stable strategy for RDEPENDS that doesn't lead to conflicts
 2) Sending request upstream to rosbridge find out why bson is listed as build dependency
 3) Adding necessary python RDEPENDS to make rosbridge stable
 4) Sending these depends upstream to rosbridge. However, I'm not sure in which extend non-ros dependencies are to be added to package.xml

bulwahn:
Okay, I agree; let us use python-pymongo for now. You can continue to discuss with the rosbridge maintainers to re-work the package.xml file.

Thanks for the other remark on how to redesign the layer to have indigo and kinetic in one branch. I will keep this in mind when we start gathering experience on maintenance of multiple ROS distribution layers.

So, the python-pymongo dependency is resolved.
What's the status on python3 recipes and should we drop python-six as dependency?
@johannesschrimpf Will you provide the commits? Then we will test them and merge this pull request.

shr-project:
Even with latest master of meta-ros and meta-python each build is showing extra note:
  NOTE: Multiple providers are available for runtime python-bson (python-bson, python-pymongo)
  Consider defining a PREFERRED_RPROVIDER entry to match python-bson
I understand that you don't want to maintain too many meta-ros branches (for each Yocto release) but the backwards compatibility needs to stop somewhere instead of carrying all this burden inside meta-ros.

I would suggest documenting required python-pymongo change in README.md file and leaving the integration with pyro and older Yocto releases to meta-ros users. I will do the same with python3-pyyaml, python3-pyparsing needed for ROS2.

Yocto 2.3 Pyro and older aren't officially supported anymore, so people using these unsupported releases should expect some extra work when trying to combine them with latest/greatest branches of other layers (creating .bbappend with python-pymongo change or backporting python3-pyyaml, python3-pyparsing recipes isn't so difficult as long as README.md helps you to find out needed backports).

shr-project (dismissed):

herb-kuta-lge (approved):
I've not tested these changes.

johannesschrimpf:
I think this is still a good improvement since the python-bson recipe is not needed and should be deleted in this layer. However, for new yocto/openembedded versions (rocko+), python-bson is part of python-pymongo (runtime).

Since LAYERSERIES_COMPAT_ros-layer is set to rocko and upwards, we now actually can use python-bson from meta-python as RDEPEND. I don't use rosbridge anymore myself, so I don't have a test setup available, but I would recommend the following:

- Set RDEPENDS to python-bson (provided by meta-python python-pymongo)
- If needed set DEPENDS to python-pymongo
- Consider using PYTHON_PN to open for python3

Here the change included in the rocko recipe:
http://cgit.openembedded.org/meta-openembedded/tree/meta-python/recipes-devtools/python/python-pymongo.inc?h=rocko

shr-project (approved):

herb-kuta-lge:
@shr-project < FYI: This wasn't rebased to the current HEAD before merging.

shr-project:
@herb-kuta-lge I'm sorry about that, I have rebased it locally after merging the other PR, but must have forgotten to force push it before merging this one.

PR #511, opened 2017-05-16
rosserial-python: Added python-pyserial to RDEPENDS

johannesschrimpf:
python-pyserial is a runtime dependency of rosserial-python, but was not added yet.

### DIFF
--- a/recipes-ros/rosserial/rosserial-python_0.6.4.bb
+++ b/recipes-ros/rosserial/rosserial-python_0.6.4.bb
@@ -5,4 +5,4 @@ LIC_FILES_CHKSUM = "file://package.xml;beginline=10;endline=10;md5=d566ef916e9de
 
 require rosserial.inc
 
-RDEPENDS_${PN} = "rosserial-msgs diagnostic-msgs"
+RDEPENDS_${PN} = "rosserial-msgs diagnostic-msgs ${PYTHON_PN}-pyserial"


bulwahn:
This commit makes sense. Up to now, we have not cared to much about run-time dependencies for ROS python packages, and rarely tested the run-time dependencies for completeness. So, it is highly probable that there is much more work to make the RDEPENDS complete in a number of meta-ros recipes. Unfortunately, testing them is rather time-consuming as well; it is not just a quick test build, but you need to create a image, start the qemu emulator, set up the environment, run the python script and then check for runtime errors, which then sometimes only occur when go through the different options of that script.

That is a lot of effort, and I won't test this commit, but we can merge it assuming that you probably did this to discover the missing run-time dependency.

johannesschrimpf:
Due to the difficulty for testing on the developer level I suggest as a routine to generate/review/include RDEPENDS by examining the run_depends in package.xml and verifying the package name.

Even though hard to test, RDEPENDS are very important for distributions with package management and if they are missing it may lead to runtime error which is critical in environments where the build server is not available, which is usual for embedded systems.

rojkov:
If the package is compatible with python3 (and the recommendation for the Indigo release is that it should) then `python-pyserial` needs to be replaced with `${PYTHON_PN}-pyserial`.

Otherwise looks good to me.

johannesschrimpf:
Ok, I changed it now to depend on ${PYTHON_PN}-pyserial .

bulwahn:
It looks good to me now; we could integrate it as is, or you could simply squash the two commits together and we merge it then. I do not see much value of having two commits, one commit is good enough.

johannesschrimpf:
Ok, it's squashed now.

bulwahn:
Although I did not expect any regressions by our usual testing with `bitbake core-image-ros-world` on a clean build, I did it anyway with that build configuration:
```

Build Configuration:
BB_VERSION        = "1.33.4"
MACHINE           = "qemux86"
TARGET_FPU        = ""
meta              = "master:1d9a88f635549e68562de681e297b9270ad02d4e"
meta-oe           
meta-multimedia   
meta-python       = "master:d3dd8bc3721f3958eb7ddeb026f7221d48daa986"
meta-ros          = "BluEye-Robotics-rosserial-rdepends:9ac944fb36b9ebfdb2ac5bb362c0eda1ccd64164"
```
and `GCCVERSION = "6%"` in the `conf/local.conf`.

Result: No errors, no surprise.

PR #512, opened 2017-05-16
unique-identifier: Updated to version 1.0.6, added meta package

johannesschrimpf:
Updated to version 1.0.6

Added meta package which is also present in the unique_identifier repository.


bulwahn:
First, If there is no special reason, we try to keep the packages in line with the versions that common are in the indigo distribution. As far as I can judge, the unique_identifier package is currently distributed in indigo as version 1.0.5, which you can check in the distribution.yaml at line 14759:

https://github.com/ros/rosdistro/blob/master/indigo/distribution.yaml#L14759

There are a few packages where we moved away from that policy, but this was usually argued by users that needed certain functionality that was only provided with the new version of that package.

Second, we currently do not create recipes for these meta packages in ROS, and I do not see the large benefit of having them in meta-ros. We do have a few packagegroups with a larger set of packages, and if you have a need to create an another packagegroups for a set of packages, please go ahead and do so. The unique-identifier meta package in isolation seems to not really offer a set of new packages that conceptionally bundle a complete specific use case.

johannesschrimpf:
Regarding the meta package, I would like to point out the difference between å package group and a meta package from the ROS point of view:
A meta package is a package which has a package.xml and is registered on the ROS page, i.e. http://wiki.ros.org/unique_identifier

A reason for including these is that other packages may depend of them. If not including the meta-package, one could not just look into the package.xml for dependencies (or the used_by tag in the wiki), but one has to look in these packages in term to see what they depend on.

While I totally agree with package groups and meta packages that are not defined as ROS packages with a package.xml I see a strong benefit of keeping the dependencies close to the ROS system. This is especially important for runtime dependencies when creating distributions with package management.

In this regard I would encourage to reconsider this policy, not so much for the unique_identifier package in special since no other packages depend on it, but as a general design choice.

Considering the version number, I agree to follow the scheme in distribution.yaml. I saw in the wiki that indigo followed the master branch but did not see the lock on this version.

bulwahn:
I do not understand this:

> If not including the meta-package, one could not just look into the package.xml for dependencies (or the used_by tag in the wiki), but one has to look in these packages in term to see what they depend on.

I believe that it was decided that ROS packages shall have in the package.xml only dependencies to real ROS packages, and not to meta packages. I remember seeing this discussion and decision in the course of the topic on package.xml version 2 a couple years ago.
So, all the incidents where a package.xml refers to a meta-package are historic legacy issues.

johannesschrimpf:
What I meant is the following:

Example: Somebody generates a package which depends on for example rosbridge_suite, and lists rosbridge_suite as run_depend in the package.xml:

  <run_depend>rosbridge_suite</run_depend>

If I were to use this package, I would expect that rosbridge-suite is present as package in yocto/meta-ros as it is in the ROS-world. Since this is not the case, I would have to figure out manually what dependencies rosbridge_suite has by looking in the package.xml of rosbridge_suite. There I would find:

  <run_depend>rosbridge_library</run_depend>
  <run_depend>rosbridge_server</run_depend>
  <run_depend>rosapi</run_depend>

Now I would have to include these in my recipe dependency list instead of just adding rosbridge-suite as dependency.

Even though the meta-packages are not supposed to be a dependency in a package.xml, according to the decision you mention, I see that that may happen.

I'm okay with the decision to omit meta-packages even though I personally see benefits in keeping the structure as most as possible like in the ros package system, which includes these meta packages.

bulwahn:
I understand your point, but I would prefer to first identify those packages in indigo that list an dependency on a meta-package, rather than adding a lot of recipes that are neither used, nor tested.

If you can name a package that has that dependency and we have a recipe for that package in meta-ros, we can add a recipe for the meta package. Otherwise, I would suggest to close this pull request.

johannesschrimpf:
My package uses the real ROS packages and the ROS wiki does not list any packages that depend on this meta package. So it's okay for me to close it.

PR #515, opened 2017-05-25
log4cxx doesn't support http, only https, therefore, the recipe failed

sebastiengemme:
Signed-off-by: Sebastien Gemme <sebastien.gemme@canada.ca>


sebastiengemme:
Recipe for log4cxx was failing, complaining about TLS errors. I changed the protocol from http to https in the recipe (and it compiled successfully).

rojkov:
Looks good to me and seems to work for me too. Thanks!

PR #517, opened 2017-06-02
yaml-cpp: disable test & util

dominiquehunziker:
Both the tests and utilites are neither run nor installed.

### DIFF
--- a/recipes-devtools/yaml-cpp/yaml-cpp_0.5.3.bb
+++ b/recipes-devtools/yaml-cpp/yaml-cpp_0.5.3.bb
@@ -14,6 +14,6 @@ SRC_URI = "https://github.com/jbeder/yaml-cpp/archive/release-${PV}.tar.gz"
 SRC_URI[md5sum] = "64200ca0bf5e0af065804d8d3e8f6d42"
 SRC_URI[sha256sum] = "ac50a27a201d16dc69a881b80ad39a7be66c4d755eda1f76c3a68781b922af8f"
 
-EXTRA_OECMAKE = "-DBUILD_SHARED_LIBS=ON"
+EXTRA_OECMAKE = "-DYAML_CPP_BUILD_TOOLS=OFF -DBUILD_SHARED_LIBS=ON"
 
 inherit cmake


bulwahn:
The change of the commit itself looks good to me. I have also successfully tested the commit on this build configuration:
```
Build Configuration:
BB_VERSION        = "1.33.4"
MACHINE           = "qemux86"
TARGET_FPU        = ""
meta              = "master:bafb7e221d40d7a87a02cec8a97d98eec7c23438"
meta-oe           
meta-multimedia   
meta-python       = "master:b063789560bfb9c60a7a15277b5b3a9839b5ba74"
meta-ros          = "rapyuta-disable_test:658dad33c68876572dbeb8645ed5739350a44854"
```
However, I have some remarks on the commit message and the context and intention (first the possibly trivial points, then the harder ones):
  - The commit itself has two different emails for the author and the committer. Is there a specific reason for using two different emails in those different roles, or is just a coincidence due to use of different machines with different default git user configurations? I would prefer to just have one email, which can be used to reach the author in case of questions. If you need help to change this with git, just let me know.
  - The commit message contains a typo, _utilites_ should be _utilities_.
  - In general, I would like to understand and the commit message should explain the following points: What is the reason for this commit (reducing image size, reducing dependencies, avoiding some runtime failure)? How was this discovered, e.g., some review of the artefacts, some failure, another activity for another specific goal? Why is the yaml-cpp recipe modified, are there possibly more recipes that will be addressed in future with the same intention? What are the drawbacks of removing the tests and utilities, will this deactivate some functionality on the existing image?

dominiquehunziker:
I modified the author and commit message. Let me know if this is ok. As for the reason why I started to look into this recipe was that it did not build for me due to an error with gmock. Disabling the test & util fixes this and I believe this is an issue with my particular setup and unrelated to the recipe itself.

bulwahn:
Maybe you can simply add the error message that occurred with gmock in your build setup, and a comment, such as: _In my particular setup, i.e., with gmock installed on the host system/image (?), building yaml-cpp failed with: ... Rather than addressing the issue in yaml-cpp, this commit deactivates the tests and utilities that fail with the setup. As both the tests and utilities are not installed and not packaged, this commit should not cause any build failures for any other unrelated setup._

Does this commit message reflect your situation and why you came up with that patch?

dominiquehunziker:
Unfortunately, I'm unable to reproduce the error (I no longer have access to the setup which produced this error). So, I'm not longer able to say what exactly the error was.

bulwahn:
So let us document that state in the commit message with: _In a specific build setup that we are not able to reproduce anymore, building yaml-cpp failed when compiling the utilities and tests. Rather than addressing the issue in yaml-cpp, this commit deactivates the tests and utilities that failed with that setup. As both the tests and utilities are not installed and not packaged, this commit should not cause any build failures for any other unrelated setup._

bulwahn:
When we then potentially revisit this commit in a couple years, this commit message will clearly point out that its intention was to improve the situation for a very specific build setup, while hoping not to break all other existing build setups.

herb-kuta-lge:
Change already appears in commit d17b8e1e055d89c4ebd0cf5f3045ee5abcb43971.

PR #537, opened 2017-08-01
[TEST] this is CI test

bbcibot:
please ignore


rojkov:
The test build failed as expected (https://bbci.blob.core.windows.net/logs/boiling-fall), but the machinery was too eager to kill the VM instance before it had a chance to report the build status. I'll fix that and retrigger the build.

rojkov:
Actually the status was reported successfully, but in a wrong cloned repository. I mistakenly infer target repository from the token's owner...

rojkov:
Both success and failure cases seem to work now.

PR #545, opened 2017-08-25
New devtool plugin for generating catkin recipes

mdhorn:
Currently the plugin only allow for processing a single
ROS package in a repository. If a repository contains more than
one package, or the package is not in the root of the repository,
then use the `--src-subdir=<dir>` option.

ROS repositories generally do not use `master` as their default
branch, so be sure to include the correct branch for the desired
distribution as part of the URI: `<URI>;branch=indigo-devel`

### DIFF
--- a/lib/recipetool/README.md
+++ b/lib/recipetool/README.md
@@ -1,0 +1,51 @@
+Extension to recipetool to enable automatic creation of
+BitBake recipe files for ROS packages.
+
+## REQUIREMENTS ##
+
+Two changes to the recipetool are required to use this plugin:
+
+Requires either:
+    From OE-Core rev: 1df60b09f7a60427795ec828c9c7180e4e52f98c
+    From OE-Core rev: 3bb979c13463705c4db6c59034661c4cd8100756
+or
+    From poky rev: b1f237ebd0d4180c5d23a0ecd9aaf7193c63a48a
+    From poky rev: a7baa47c876c7895499909731aaa451c6009610a
+
+These changes are currently in the master branch as of 2017-Aug-24.
+
+
+## USAGE ##
+
+  Initialize the build environment:
+
+    source oe-init-build-env
+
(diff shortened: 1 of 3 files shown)


mdhorn:
retest this please

rojkov:
@mdhorn Thanks for the PR. Currently the test builds fail because of https://github.com/bmwcarit/meta-ros/issues/542. It's not related to the devtool plugin. The code itself looks very good and clean.
I'd rather use string extrapolation like
```
'RDEPENDS_${PN}-dev += "%s"' % dependency
```
instead of
```
"RDEPENDS_${PN}-dev += \"" +  dependency + "\"
```
but that's just a nitpick.

mdhorn:
@rojkov Please let me know if you are okay with merging this or if you'd like me to make the change to the string extrapolation.

rojkov:
The plugin works as expected and I'm ok with the current code. Will merge it as soon as CI gets green (hopefully after merging https://github.com/bmwcarit/meta-ros/pull/546).

bulwahn:
I had a first look at the tool looks well developed with some minimal documentation in place etc. So, this looks very good to me.
I will also review in more depth and test the tool (hopefully this week or next week); possibly, I also have some time to write some first sanity tests to check if the tool in principle still works with the bitbake version in the CI that will update over time. All of that should not stop us to merge the pull request now. Another topic I noticed is that it licensed under GPLv2, which makes sense as it is an extension to bitbake. It might make sense to point out in the README that the tool is not MIT-licensed as the recipe, but GPLv2.

mdhorn:
@bulwahn Are the README changes something you want me to make to enable merging this PR?

bulwahn:
The README changes would be nice, but are not a blocker for merging this pull request. I just did not have time to test the script yet, and probably I will not have time in the close future, so we can go ahead and merge anyway.

PR #556, opened 2017-10-06
simple-message: fix up patch's upstream status

rojkov:
And the corresponding kinetic-devel version is https://github.com/ros-industrial/industrial_core/pull/184.


bulwahn:
Thanks, I will merge it without further ado.

PR #557, opened 2017-10-19
tf2: ship tf2-msgs with tf2

Minipada:
## What I did
I added `tf2-msgs` as a run dependency to `tf2`.

## Why
Since tf2-msgs was not a run dependency, it was not shipped when only using `tf2`, creating runtime issues. I checked the manifest, it is now included properly

## Versions I use
meta-ros@88ce154 (12th June 2017)
poky 2.2

### DIFF
--- a/recipes-ros/geometry2/tf2_0.5.16.bb
+++ b/recipes-ros/geometry2/tf2_0.5.16.bb
@@ -6,4 +6,6 @@ LIC_FILES_CHKSUM = "file://package.xml;beginline=16;endline=16;md5=d566ef916e9de
 
 DEPENDS = "tf2-msgs geometry-msgs console-bridge rospy"
 
+RDEPENDS_${PN} += "tf2-msgs geometry-msgs console-bridge rostime"
+
 require geometry2.inc


rojkov:
The first CI build failed because of reaching disk limit for bitbake's build artifacts. I've doubled the limit and restarted the build.

As bitbake's cache is reused from the previous build this one should complete faster.

bulwahn:
@Minipada Thanks for that patch. Just two easy improvements and one question:
- I think you can simply put the RDEPENDS in one line; there is no need for the line breaks (and it remains comprehensive even in one line when a few more run-time dependencies are added).
- The commit message names the recipe before the colon and we use the imperative (now), so it is `ship` not `ships`, so overall, the commit message is `tf2: ship tf2-msgs with tf2` if you want to keep it as is; otherwise `tf2: improve runtime dependency` is also fine.
- It also depends on geometry-msgs; do we need to add that as a runtime dependency as well?

When two points above are addressed, I will merge it.

Minipada:
@bulwahn My pleasure, I have addressed the two first points.

About the third, I have added all of them following the [package.xml](https://github.com/ros/geometry2/blob/indigo-devel/tf2/package.xml). I personally rely a lot on `RDEPENDS` to have minimalist images containing only project packages, which themselves rely on standard layer like this one

bulwahn:
Thanks, that looks good to me now. I will merge it.

PR #568, opened 2017-12-11
gitignore: Ignore vim swp files

johannesschrimpf:
Add `.*.swp` to .gitgnore to ignore vim swp files


bulwahn:
Okay, that looks unproblematic; I will merge it.

PR #585, opened 2018-04-06
layer.conf: add LAYERSERIES_COMPAT

shr-project:


bulwahn:
I am happy to include it. Could you say a bit more on its motivation though?

shr-project:
Hi Lukas,

read http://lists.openembedded.org/pipermail/openembedded-core/2018-April/149705.html and you can see warning if you try to use meta-ros without this change with today's version of oe-core.

bulwahn:
Thanks, that will do. I will have a look, let our CI run through and then merge it.

PR #590, opened 2018-04-15
rebased pull request #519

bulwahn:
This pull request is a rebase of pull request #519. It adds support for the SDK feature and allows users to compile on the target system.


bulwahn:
This pull request will be merged in two weeks, if I don't get any user feedback that it does not work or breaks something.

bulwahn:
@rojkov the CI failed with:
```
WARNING: The free space of /fileshare/bb-cache/downloads (//bmwcarityottacigh.file.core.windows.net/bbcache) is running low (0.000GB left)
[...]
Summary: There was 1 ERROR message shown, returning a non-zero exit code.
ERROR: Immediately abort since the disk space monitor action is "ABORT"!
```
What can we do here?

rojkov:
That was another glitch in Azure: for a short moment the file storage with downloaded sources started to report 0 free space even though the current consumption is 33G of 5T limit. Probably the cluster machine serving the storage rebooted. I've restarted the build.

recipes-devtools/python/python-catkin-pkg.inc
```diff
@@ -9,4 +9,7 @@ SRC_URI[sha256sum] = "d4d069eaeb29d6a9ee71773b20c41515b6e0feb9ea2765ebaf3e599d0c
 
 inherit pypi
 
-BBCLASSEXTEND = "native"
+RDEPENDS_${PN} = "${PYTHON_PN}-unixadmin"
+RDEPENDS_${PN}_class-native = ""
+
+BBCLASSEXTEND = "native nativesdk"
```
Minipada: Don't we need another line containing

Minipada: Don't we need another line under it:FILES_${PN}_append_class-nativesdk = " ${SDKPATHNATIVE}"

Minipada, line 15: Don't we need another line under this

    FILES_${PN}_append_class-nativesdk = " ${SDKPATHNATIVE}"

recipes-ros/catkin/catkin-runtime_0.6.19.bb
```diff
@@ -26,6 +27,27 @@ do_install_append() {
...
+}
+
+# Append environment hook for SDK
+do_install_append_class-nativesdk() {
+    mkdir -p ${D}${SDKPATHNATIVE}/environment-setup.d
+
+    script=${D}${SDKPATHNATIVE}/environment-setup.d/catkin-runtime.sh
+    echo "export ROSDISTRO=${ROSDISTRO}" >> $script
+    echo "export PATH=\$PATH:\${OECORE_NATIVE_SYSROOT}/opt/ros/\${ROSDISTRO}/bin" >> $script
+    echo "export PYTHONPATH=\${OECORE_NATIVE_SYSROOT}/opt/ros/\${ROSDISTRO}/lib/python2.7/site-packages" >> $script
```
Minipada, line 42: Don't we also need?

    \${OECORE_TARGET_SYSROOT}/usr/lib/python2.7/site-packages

Minipada (changes requested):
Those are general comments for now, from differences I noticed with my local SDK branch. 

I started building and required to add cmake-modules.

    BBCLASSEXTEND = "native nativesdk"

I'm now building it and will come back when I the SDK generation is done

recipes-devtools/python/python-catkin-pkg.inc
bulwahn: @Minipada I have just rebased the commits from @dominiquehunziker, and I did not test the SDK feature. So, I cannot really comment on what is needed and what is not. If you think it is needed, I would cherry-pick patches that you provide to this pull request.

recipes-ros/catkin/catkin-runtime_0.6.19.bb
bulwahn: I will add it if you find out that it is needed.

otavio: This shouldn't be needed; if it is, it is an indication something is wrong as this should work out of box.

recipes-devtools/python/python-catkin-pkg.inc
Minipada: In case you didn't notice, I made a PR on your fork

dominiquehunziker:
I wanted to clarify the issue of the Python2/3 support. Is this addressed now or is the SDK limited to Python2 for the time being? 

Unfortunately, I no longer work with open-embedded directly. So, it is harder to find the time to try new things.

bulwahn:
We resolved the python2/3 support problem with the commit https://github.com/bmwcarit/meta-ros/commit/6ddc0cc518ebcfdd6871774030bb5f91c09557d9. Since then, the patches of your SDK support still allow to cross-compile with python2 and python3, but I have not tested if the SDK feature itself works. Hopefully, @Minipada and others can test that and report if the SDK feature works, potentially with both setups, python2 and python3.

recipes-ros/catkin/catkin-runtime_0.6.19.bb
dominiquehunziker, line 42: In this line Python2.7 is hard-coded. This should depend on the version installed in the SDK (native).

recipes-ros/ros/files/0001-use-env-to-get-python-exec.patch
```diff
@@ -0,0 +1,32 @@
...
+diff --git a/core/roslib/env-hooks/10.ros.sh.em b/core/roslib/env-hooks/10.ros.sh.em
+index caed342..1c64fc4 100644
+--- a/core/roslib/env-hooks/10.ros.sh.em
++++ b/core/roslib/env-hooks/10.ros.sh.em
+@@ -1,7 +1,7 @@
+ # generated from ros/env-hooks/10.ros.sh.em
+ 
+ # scrub old ROS bin dirs, to avoid accidentally finding the wrong executables
+-export PATH="`@(PYTHON_EXECUTABLE) -c \"import os; print(os.pathsep.join([x for x in \\\"$PATH\\\".split(os.pathsep) if not any([d for d in ['cturtle', 'diamondback', 'electric', 'fuerte'] if d in x])]))\"`"
++export PATH="`/usr/bin/env python -c \"import os; print(os.pathsep.join([x for x in \\\"$PATH\\\".split(os.pathsep) if not any([d for d in ['cturtle', 'diamondback', 'electric', 'fuerte'] if d in x])]))\"`"
```
dominiquehunziker, line 20: This should be fine, but maybe better to be explicit about which version to use.

recipes-ros/ros/files/0001-use-env-to-get-python-exec.patch
```diff
@@ -0,0 +1,32 @@
...
+ 
+ if [ -n "$ROS_DISTRO" -a "$ROS_DISTRO" != "indigo" ]; then
+   echo "ROS_DISTRO was set to '$ROS_DISTRO' before. Please make sure that the environment does not mix paths from different distributions."
+@@ -30,7 +30,7 @@ for workspace in workspaces:
+ print(os.pathsep.join(paths))
+ EOF
+ )
+-export ROS_PACKAGE_PATH="`@(PYTHON_EXECUTABLE) -c \"$PYTHON_CODE_BUILD_ROS_PACKAGE_PATH\"`"
++export ROS_PACKAGE_PATH="`/usr/bin/env python -c \"$PYTHON_CODE_BUILD_ROS_PACKAGE_PATH\"`"
```
dominiquehunziker, line 29: same here

dominiquehunziker (commented):
I marked some lines which IMO are going to be problematic if Python2/3 is going to be supported.

dominiquehunziker:
I tried to run through the steps from [here](https://github.com/bmwcarit/meta-ros/pull/492#issuecomment-305955808); however, I encountered some issues. I hope I will have some time to formalize the steps a bit more in order to reproduce them more easily. Nevertheless, here the issues encountered so far:
1) `nativesdk-cmake` is missing a file (instead it's placed in `nativesdk-catkin-dev`). IMO this is a bug in the catkin recipe as the missing file is required by a script that is placed in `nativesdk-cmake`. Found the following [SO question](https://stackoverflow.com/questions/49532977/yocto-generated-nativesdk-cmake-sdk-is-incomplete) which proposes one solution. My proposal is to apply the following patch to the cmake recipe
```diff
diff --git a/meta/recipes-devtools/cmake/cmake_3.10.3.bb b/meta/recipes-devtools/cmake/cmake_3.10.3.bb
index 3f8fd7a..edffa4f 100644
--- a/meta/recipes-devtools/cmake/cmake_3.10.3.bb
+++ b/meta/recipes-devtools/cmake/cmake_3.10.3.bb
@@ -43,6 +43,7 @@ do_install_append_class-nativesdk() {
 }
 
 FILES_${PN}_append_class-nativesdk = " ${SDKPATHNATIVE}"
+FILES_${PN}-dev_class-nativesdk = ""
 
 FILES_${PN} += "${datadir}/cmake-${CMAKE_MAJOR_VERSION}"
 FILES_${PN}-doc += "${docdir}/cmake-${CMAKE_MAJOR_VERSION}"
```

2) I noticed that both Python2 & Python3 are being built and installed in the image; not sure why this is. In it's current form though at least 1 dependency for catkin is missing in the target image when building with the default settings; namely Python's `argparse`. Interestingly, `argparse` is available in Python3 but not in Python2 which looks like is the default.

As far as I can tell the SDK works fine -- It was able to compile (the very simple) ROS package (from the previously linked steps); however, I was not able yet to check whether the binaries work correctly.

bulwahn:
@dominiquehunziker Thanks for the testing effort. I will include the proposed patch; could you provide a patch to meta-ros on your fork and send me a link to it and then I will cherry-pick it into this pull request?

dominiquehunziker:
So, I was able to run through my steps and I formalized it a bit [here](https://github.com/dominiquehunziker/meta-ros-test-sdk). Additionally, I pushed the changes [here](https://github.com/dominiquehunziker/meta-ros/commits/rebased-pull-request-519).
However, due to some very recent changes to openembedded-core (see [this issue](https://github.com/bmwcarit/meta-ros/issues/592)), I was not able to test everything from a clean setup. Also, I only tested the Python2 version.

bulwahn:
@dominiquehunziker I will have a look at the further changes that you propose. Then I will test the basics once again, and then we will consider merging it, although the sdk support might now just only work with python. Concerning openembedded-core changing while you try to resolve a complex build issue, I know just too well. Once you are close to done with your feature, you need to hunt down some new bug to actually test the new feature on the HEAD commit. I will try to resolve the issues with the new boost version. If you have time to test the sdk feature once again, just use the openembedded-core commit before the boost update and then report your finding. I assume that the sdk feature is independent of this boost update commit: so if the sdk works on the previous commit, it will probably still work afterwards once we have merged patches for the boost update.

dominiquehunziker:
@bulwahn I realized for the packages that I need for running the SDK test there are not that many changes necessary; so, I I quickly added a commit with some patches. I will leave it up to you on how you want to handle this.
As for the tests: I was able to run through them again with the script I linked [here](https://github.com/dominiquehunziker/meta-ros-test-sdk). The script unfortunately is not completely stand-alone (didn't find a nice solution to shutdown qemu from the script).

flplv:
Tested @dominiquehunziker test script, worked seamlessly. It cross-build fine via the SDK and it also built locally in the qemu-arm too. 
Also tested cross-building locally (instead of docker) using the @dominiquehunziker test script SDK, worked. 
I did not test the catkin support for unit test build though.

bulwahn:
@Minipada @dominiquehunziker @flplv Thanks to all of you for testing. I included @dominiquehunziker  's patches in this pull request, and tested `bitbake core-image-ros-world` successfully with and without `ROS_USE_PYTHON3` on this build configuration:

```
Build Configuration:
BB_VERSION           = "1.37.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "ubuntu-16.04"
TARGET_SYS           = "i586-oe-linux"
MACHINE              = "qemux86"
DISTRO               = "nodistro"
DISTRO_VERSION       = "nodistro.0"
TUNE_FEATURES        = "m32 i586"
TARGET_FPU           = ""
meta                 = "revert-boost-update:e213f09799d887bd0a1ca6aa20daaf52546a7f2c"
meta-oe              
meta-python          
meta-multimedia      = "master:a19aa29f7fa336cd075b72c496fe1102e6e5422b"
meta-ros             = "rebased-pull-request-519-plus-additions:808dcb3a1ed8cb63784a3a972c01d36267fae54c"
```

The commit `e213f09799d887bd0a1ca6aa20daaf52546a7f2c` is`f04e6bd144deb0c8fe2742f66b18904b6619a502` + reverting `358cf46ea4d01b7ad8c355fa103d4a6922cc0a88 ("boost: update to 1.67.0")`.

I see that the travis-ci reports missing Upstream-Status, but I expect that we will fix this when we proper fixes for the boost update issues.

@Minipada I had a look at https://github.com/bulwahn/meta-ros/pull/13, and I decided to not include it in this pull request (which is already quite big). You can open a pull request for that new feature, once this feature is merged.

flplv:
Also tried build the @dominiquehunziker test with this environment plus the nodelet recipe active, but failed with Boost error:
```
Build Configuration:
BB_VERSION           = "1.37.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "ubuntu-16.04"
TARGET_SYS           = "arm-oe-linux-gnueabi"
MACHINE              = "qemuarm"
DISTRO               = "nodistro"
DISTRO_VERSION       = "nodistro.0"
TUNE_FEATURES        = "arm armv5 thumb dsp"
TARGET_FPU           = "soft"
meta                 = "master:70c1154163761253346fb477ff362af6a838be09"
meta-oe              
meta-python          
meta-multimedia      = "master:a19aa29f7fa336cd075b72c496fe1102e6e5422b"
meta-ros             = "master:8a9faf90c6e1145770f44653efc635c05971cecf" <---- CHANGED TO MASTER
```
error:
```
 error: no matching function for call to 'boost::date_time::subsecond_duration<boost::posix_time::time_duration, 1000>::subsecond_duration(double)'
|      condition_.timed_wait(mutex_, boost::posix_time::milliseconds(wait_time.toSec() * 1000.0f));
|                                                                                               ^
```

dominiquehunziker:
@flplv 

Regarding the hardcoded paths: In my case there are:
* non ROS related (added some comments for completeness; however, these are unrelated to the meta-ros SDK stuff; so if there is an issue it would more likely be on the openembedded layer)
    * `sysroots/x86_64-oesdk-linux/usr/bin/x86_64-oesdk-linux-libtool` and `sysroots/armv5e-oe-linux-gnueabi/etc/opkg.conf` which I will leave up to the openembedded mailing-list to answer; best guess though is that they are irrelevant;
    * a bunch of `*/.debug/*` files which are just debug symbols for gdb;
    * a bunch of `*.pyc` files where the paths are used for debug information;
    * `sysroots/x86_64-oesdk-linux/usr/lib/arm-oe-linux-gnueabi/gcc/arm-oe-linux-gnueabi/7.3.0/plugin/include/configargs.h` which are the arguments that have been used to build gcc and are for information purposes only;
* ROS related
    * `pkgConfig.cmake` files (such as `sysroots/armv5e-oe-linux-gnueabi/opt/ros/indigo/share/rosmaster/cmake/rosmasterConfig.cmake`): This is due to some catkin "magic". If you open the files you notice that the variables containing the hardcoded paths are never used (has something to do with devel-/installspace);
    * message header files (such as `sysroots/armv5e-oe-linux-gnueabi/opt/ros/indigo/include/rosgraph_msgs/Log.h`): You notice this is actually part of a comment in the header file;
    * pkg-config files (such as `sysroots/armv5e-oe-linux-gnueabi/opt/ros/indigo/lib/pkgconfig/rostime.pc`): Indeed these types of files are wrong. Though that is also the case for regular target builds; not just for the SDK. Given that nobody fixed this yet though probably means that only very few people are using pkg-config in conjunction with ROS packages instead of catkin.

Regarding `kinetic` build: This is possible. I only tested and verified one setup (and a rather simple one at that) and only for indigo (as this is the currently maintained release here as far as I'm aware). Especially with the SDK the number of possible combinations that could be tested increases dramatically.

Regarding the Boost error: Seems to be related to #592

flplv:
Thanks a lot @dominiquehunziker. You helped me to isolate the problem.
Seems to be either:
a) `nodeletCmake.config` problem
b) my recipes

Since in my Yocto project I see a different consturction of the `nodeletCmake.config` from `roscppCmake.config` inside the sdk:

ROSCPP:
```
if(NOT "include;/usr/include " STREQUAL " ")
  set(roscpp_INCLUDE_DIRS "")
  set(_include_dirs "include;/usr/include")
```

NODELET:
```
if(NOT "include;/home/arczi/work/Aeolus/yocto/build/tmp/work/cortexa17hf-neon-vfpv4-poky-linux-gnueabi/ros-nodelet/1.9.5+gitAUTOINC+2ec24ff596-r0/recipe-sysroot/usr/include " STREQUAL " ")
  set(nodelet_INCLUDE_DIRS "")
  set(_include_dirs "include;/home/arczi/work/Aeolus/yocto/build/tmp/work/cortexa17hf-neon-vfpv4-poky-linux-gnueabi/ros-nodelet/1.9.5+gitAUTOINC+2ec24ff596-r0/recipe-sysroot/usr/include")
```

This is the exact problem. If I could compile the sdk-test app with the nodelet (boost problem) I would be able to isolate a) and b).

n0rthernlight:
Hello @dominiquehunziker. I made some investigation regarding the problem and it looks that the issue is strictly related to build/catkin_generated/installspace.
For **ROSCPP** package the `roscppConfig.cmake` is generated with:

```
if(NOT "include;/usr/include " STREQUAL " ")
  set(roscpp_INCLUDE_DIRS "")`
  set(_include_dirs "include;/usr/include")
```

which is correct, then for **NODELET** package (`nodeletConfig.cmake`) it's hardcoded:

```
if(NOT "include;/home/arczi/work/Aeolus/yocto/build/tmp/work/cortexa17hf-neon-vfpv4-poky-linux-gnueabi/ros-nodelet/1.9.5+gitAUTOINC+2ec24ff596-r0/recipe-sysroot/usr/include " STREQUAL " ")`
  set(nodelet_INCLUDE_DIRS "")
  set(_include_dirs "include;/home/arczi/work/Aeolus/yocto/build/tmp/work/cortexa17hf-neon-vfpv4-poky-linux-gnueabi/ros-nodelet/1.9.5+gitAUTOINC+2ec24ff596-r0/recipe-sysroot/usr/include")
```

Which is causing issues. Is there any quick solution here how to get rid of this? Might be there something extra in ROS meta-layer that omits that hardcoding?

flplv:
Guys, @n0rthernlight and I are working in the same issue together. 
The problem with the nodelet was internal, we had a changed version of the nodelet_xxx.bb file. So, in my previous message it was b).
Sorry for bothering you guys, thanks a lot for the support @dominiquehunziker.

PR #597, opened 2018-05-24
Fix installed but not not shipped error

absci:
6 installed and not shipped files.
  /usr/lib/cmake
  /usr/lib/cmake/yaml-cpp
  /usr/lib/cmake/yaml-cpp/yaml-cpp-config.cmake
  /usr/lib/cmake/yaml-cpp/yaml-cpp-config-version.cmake
  /usr/lib/cmake/yaml-cpp/yaml-cpp-targets-release.cmake
  /usr/lib/cmake/yaml-cpp/yaml-cpp-targets.cmake
Add those files to FILES_${PN}

Bitbake error:
```
ERROR: yaml-cpp-0.6.2-r0 do_package: QA Issue: yaml-cpp: Files/directories were installed but not shipped in any package:
  /usr/lib/cmake
  /usr/lib/cmake/yaml-cpp
  /usr/lib/cmake/yaml-cpp/yaml-cpp-config.cmake
  /usr/lib/cmake/yaml-cpp/yaml-cpp-config-version.cmake                                                                           
  /usr/lib/cmake/yaml-cpp/yaml-cpp-targets-release.cmake                                                                          
  /usr/lib/cmake/yaml-cpp/yaml-cpp-targets.cmake                                                                                  
Please set FILES such that these items are packaged. Alternatively if they are unneeded, avoid installing them or delete them within do_install.                                                                                                                    
yaml-cpp: 6 installed and not shipped files. [installed-vs-shipped]                                                               
ERROR: yaml-cpp-0.6.2-r0 do_package: Fatal QA errors found, failing task.
ERROR: yaml-cpp-0.6.2-r0 do_package: Function failed: do_package
ERROR: Logfile of failure stored in: /home/user/yocto/build/tmp/work/cortexa8hf-neon-poky-linux-gnueabi/yaml-cpp/0.6.2-r0/temp/log.do_package.12192
ERROR: Task (/home/user/yocto/poky/meta-ros/recipes-devtools/yaml-cpp/yaml-cpp_0.6.2.bb:do_package) failed with exit code '1'
```


bulwahn:
Thanks for this commit. I am currently on vacation and cannot test it until I am back from vacation. It generally looks okay to me, but the commit message in openembedded/yocto layers usually name the recipe in their commit message first. So it should start with yml-cpp, i.e., yaml-cpp: include non-shipped files to default package. Also why do you believe these files should be added to the default package, I would expect those to be part of the -dev package?

absci:
Thanks for the suggestion. Since the files are all there but not shipped in the package, it just can't pass the QA test in bitbake. So I added those files.

bulwahn:
I just build yaml-cpp on the current master and the issue does not occur there. I believe that is due to this recent commit in the cmake.bbclass:

http://cgit.openembedded.org/openembedded-core/commit/meta/classes/cmake.bbclass?id=d91dc4666683a96e9d03cbbd21b8a546f9069c93

@absci Which build configuration do you use?

So for the current master, we do not require the proposed commit. I would certainly propose to change the commit to be compatible to what has been changed on master, and note that this is a change solely for the support of a specific older build configuration (some poky release?).

shr-project:
agreed with @bulwahn , it doesn't belong into $PN and it should be added to $PN with:
FILES_${PN}-dev += "${libdir}/cmake"
like newer oe-core is doing, that way yaml-cpp stays compatible with current master as well older oe-core branches.

I also have small fix for packaging gtest files:
https://github.com/shr-project/meta-ros/commit/ba924902e7c216f6162a08002b9bedddf4bfb254

PR #605, opened 2018-07-13
drop patches not needed after libeigen update

bulwahn:


bulwahn:
CI is green; so, I will merge this.

Issue #613, opened 2018-11-02
bitbake yaml-cpp error


qeiottech:
ERROR: yaml-cpp-0.6.2-r0 do_package: QA Issue: yaml-cpp: Files/directories were installed but not shipped in any package:
  /usr/lib/cmake
  /usr/lib/cmake/yaml-cpp
  /usr/lib/cmake/yaml-cpp/yaml-cpp-config.cmake
  /usr/lib/cmake/yaml-cpp/yaml-cpp-config-version.cmake
  /usr/lib/cmake/yaml-cpp/yaml-cpp-targets-release.cmake
  /usr/lib/cmake/yaml-cpp/yaml-cpp-targets.cmake
Please set FILES such that these items are packaged. Alternatively if they are unneeded, avoid installing them or delete them within do_install.
yaml-cpp: 6 installed and not shipped files. [installed-vs-shipped]
ERROR: yaml-cpp-0.6.2-r0 do_package: Fatal QA errors found, failing task.
ERROR: yaml-cpp-0.6.2-r0 do_package: Function failed: do_package
ERROR: Logfile of failure stored in: /home/jaloo/backup/s32v234/src/fsl-auto-yocto-bsp/build_s32v234evb_vsdk/tmp/work/aarch64-fsl-linux/yaml-cpp/0.6.2-r0/temp/log.do_package.25518
ERROR: Task (/home/jaloo/backup/s32v234/src/fsl-auto-yocto-bsp/sources/meta-ros/recipes-devtools/yaml-cpp/yaml-cpp_0.6.2.bb:do_package) failed with exit code '1

shr-project:
You're probably using old oe-core. /usr/lib/cmake is automatically packaged to PN-dev with recent oe-core releases.

aizelauna:
To workaround this issue, you can create the file _recipes-devtools/yaml-cpp/yaml-cpp_0.6.2.bbappend_ in your layer with this content:
`FILES_${PN}-dev += "/usr/lib/cmake/yaml-cpp"`

herb-kuta-lge:
No longer happens in the [thud-draft] branch. @jaloo2018 < Can this issue be closed?

herb-kuta-lge:
"Silence implies consent" (see http://wiki.c2.com/?SilenceImpliesConsent).

PR #621, opened 2019-02-27
tf, tf2, message-filters, roscpp: remove signals from find_package() calls to support boost 1.69

### LINKED ISSUE #618: roscpp, boost 1.69 and warrior
Greetings,

boost 1.69 has got it to thud branch of oe-core, so roscpp is broken now (see https://github.com/bmwcarit/meta-ros/issues/615). Is there any chance compatibility with thud can be restored, or do we have to downgrade to sumo now?

Thanks!

UPD: looks like dependency on boost::signals has been removed shortly from ros_comm/roscpp, but it hasn't been released yet, and meta-ros seems to be designed to use release tarballs. Waiting for a release then?

aizelauna:
boost has been updated in Yocto thud branch to 1.69. This boost update removes the deprecated library boost::signals. 
Since several versions, it is replaced by boost::signals2 which is a header only library.
Ros code already uses signals2 library but it still has links to the older library in its CMake files. This commit removes those references as they are now useless and more importantly source of error at configure time.

This PR should fix #618  issue.


bulwahn:
Patch generally looks okay. I am just wondering:
  - is it possibly simpler to just update those packages? The upstream status suggests we are just lacking updates, right? I have not been updating the indigo packages much lately.
  - is this commit backwards compatible? Does this commit also work with a build configuration before boost was updated to 1.69?

aizelauna:
First or all, thanks very much for looking at my PR.

> is it possibly simpler to just update those packages? The upstream status suggests we are just lacking updates, right? I have not been updating the indigo packages much lately.

To set the status, I used the same status than this patch that adds recent boost support:
https://github.com/bmwcarit/meta-ros/blob/master/recipes-ros/ros-comm/roscpp/0001-make-implicit-cast-explicit.patch
It looks like that the decision has been taken to not update kinetic and indigo packages.

> is this commit backwards compatible? Does this commit also work with a build configuration before boost was updated to 1.69?

I'll try to test this commit with rocko branch, I'll let you know about the result.

aizelauna:
Test is going on, but I will not have any result before mid of march.
It looks like that meta-ros is not compatible with rocko, I have an issue with yaml-cpp that should be easily fixed with a bbappend (issue #613).

aizelauna:
I tested this branch of _meta-ros_ within a _rocko_ based project. Boost version is 1.64.
It works just fine. 

For the record, I had to patch _yaml-cpp_ with this line in a bbappend file (issue #613):
` FILES_${PN}-dev += "/usr/lib/cmake/yaml-cpp"`
I did not include this in the PR because this is something not required with thud.

aizelauna:
Hi,
I am coming back to get some update on this PR.
Is there anything else to do to be accepted and merged to master?

shr-project (dismissed):

shr-project (approved):

PR #626, opened 2019-04-14
Add common-interfaces itself to the 'common-interfaces' group

OYTIS:
Useful for packages that explicitly depend on common-interfaces

Signed-off-by: Anton Gerasimov <tossel@gmail.com>


shr-project:
legacy (v1) isn't maintained anymore.

PR #634, opened 2019-07-03
Various simple fixes

shr-project:


shr-project:
Needs one more fix for festival to build correctly with "speech-tools: respect LDFLAGS" change.

herb-kuta-lge (changes requested):
Won't it will confuse folks if the "legacy" branch, which isn't supposed to be under active development, appears to support a newer release of OpenEmbedded than the branch that is under active development?

herb-kuta-lge (changes requested):
Won't it will confuse folks if the "legacy" branch, which isn't supposed to be under active development, supports a newer release of pcl than the branch that is under active development?

herb-kuta-lge (dismissed):
Won't it will confuse folks if the "legacy" branch, which isn't supposed to be under active development, appears to support a newer release of OpenEmbedded than the branch that is under active development?

shr-project:
> Won't it will confuse folks if the "legacy" branch, which isn't supposed to be under active development, appears to support a newer release of OpenEmbedded than the branch that is under active development?

It's still legacy ROS :) and I hope to create a branch for warrior soon after we finalize the actively developed thud branch.

These changes at least show nicely what will need to be fixed in thud-drafts as well, when we start making it compatible with warrior (at least it will show some of it, but it will probably need changes in more generated recipes).

shr-project:
> Needs one more fix for festival to build correctly with "speech-tools: respect LDFLAGS" change.

This is now fixed as well by 1 additional commit.
In the end, the issue wasn't caused by "speech-tools: respect LDFLAGS" change, just silly way of sharing build configuration between speech-tools and festival and the issue was reproducible with or without respecting LDFLAGS.

herb-kuta-lge:
> It's still legacy ROS :) 

Perhaps a better name would have been [deprecated] :-) . We'd like everyone using `meta-ros` to start using and contributing to the version 2 layer.

> and I hope to create a branch for warrior soon after we finalize the actively developed thud branch.

Please don't do this until [thud-draft] is finalized as [master] and we have added support for all of the active ROS distros there.

herb-kuta-lge (approved):

PR #651, opened 2019-12-20
Update recipes under recipes-ros-infrastructure for new versions found in newer-platform-components.list

shr-project:


shr-project:
Merged in:
d85431585..6d4554a24  thud-draft  -> ros/thud-draft

PR #661, opened 2019-12-20
Check with yaml-cpp-vendor if yaml-cpp for ROS 2 can be migrated to v0.6.3

shr-project:


shr-project:
Resolved differently in Milestone 10.

PR #696, opened 2020-07-08
[thud] Add support for Foxy Fitzroy

shr-project:


shr-project:
Merged and tagged as thud/milestones/13 now when it's consistent with milestone 13 in other branches.

PR #698, opened 2020-07-08
[dunfell] Upgrade ROS 1 melodic to 2020-08-07 release

shr-project:


meta-ros1-melodic/recipes-bbappends/rosbag-fancy/rosbag-fancy_0.2.0-1.bbappend
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) 2020 LG Electronics, Inc.
+
+# rosbag-fancy/0.2.0-1-r0/git/src/main.cpp:11:10: fatal error: std_srvs/Trigger.h: No such file or directory
+DEPENDS += "std-srvs"
```
Rayman, line 4: Just a question, are you contributing upstream fixes for these issues? It would be good to fix these issues at the root.

shr-project: In some cases I am, depends on how much time I have and how confident I am that it's the correct fix (e.g. now there are still 10+ recipes failing for melodic so I couldn't even completely build test it all and there is already even newer melodic release to start working on).

PR #700, opened 2020-07-08
[build] Support for newer revisions of upstream OE layers

shr-project:


shr-project:
Depends on https://github.com/ros/meta-ros/pull/704

PR #704, opened 2020-07-28
[build] Add support for Foxy Fitzroy with OE thud, warrior, zeus

shr-project:


shr-project:
Merged and tagged as build/milestones/13-all, because build/milestones/13 already existed with just dunfell and gatesgarth support for foxy.

PR #722, opened 2020-08-21
[build] Add support for ROS 1 noetic and update to latest revisions of upstream layers

shr-project:


herb-kuta-lge (approved):

shr-project:
In build-next as well:
   e88327c42..26aba8735  build -> build-next

PR #739, opened 2020-09-15
[build] Support for newer revisions of upstream OE layers V

shr-project:


shr-project:
Depends on spdlog changes from https://github.com/ros/meta-ros/pull/740 (the MCF_META_ROS_COMMIT should be updated to revision after 740 is merged).

PR #747, opened 2020-10-09
[dunfell] Fix for ROS_DISTRO variable and meta-ros-python2 separation

shr-project:


meta-ros-common/conf/ros-distro/ros-distro.conf
```diff
@@ -40,7 +40,7 @@ ROS_DISTRO_METADATA_VERSION_REVISION = "${ROS_DISTRO_METADATA_MILESTONE_NUMBER}"
 ROS_DISTRO_METADATA_VERSION = "${ROS_DISTRO_METADATA_VERSION_MAJOR}.${ROS_DISTRO_METADATA_VERSION_MINOR}.${ROS_DISTRO_METADATA_VERSION_MICRO}.${ROS_DISTRO_METADATA_VERSION_UPDATE}"
 ROS_DISTRO_METADATA_VERSION .= "-${ROS_DISTRO_METADATA_VERSION_REVISION}"
 
-ROS_DISTRO ??= "${@d.getVar('ROS2_DISTRO') if d.getVar('ROS2_DISTRO') != '' else d.getVar('ROS1_DISTRO')}"
+ROS_DISTRO ??= "${@d.getVar('ROS2_DISTRO') if d.getVar('ROS2_DISTRO', True) not in (None, '') else d.getVar('ROS1_DISTRO', True)}"
```
herb-kuta-lge: Isn't `True` the default the 2nd parameter of `d.getVar()`?

meta-ros-common/conf/distro/include/ros-common.inc
```diff
@@ -1,6 +1,6 @@
 # OE DISTRO configuration settings common to both ROS 1 and ROS 2 distros.
 #
-# Copyright (c) 2019 LG Electronics, Inc.
```
herb-kuta-lge, line 3: Commit message: I've been using the convention of have a space between ROS and [12].

herb-kuta-lge (changes requested):

meta-ros-common/conf/ros-distro/ros-distro.conf
shr-project: Yes, I've added it when debugging why it didn't work and then forgot to drop it. Will do.

meta-ros-common/conf/distro/include/ros-common.inc
shr-project: in 'ROS[12]_DISTRO' when I want to express ROS1_DISTRO or ROS2_DISTRO variable?

herb-kuta-lge: No, just in the commit message, eg, 

> and enable it by default for all ROS2 distributions as well as ROS1 noetic

-> and enable it by default for all ROS 2 distributions as well as ROS 1 noetic

PR #755, opened 2020-11-03
[Draft][master] Fixes for Yocto 3.3 Hardknott

shr-project:


shr-project:
To prevent cv-bridge patch being hardknott specific I'll try to re-work it to find python_version dynamically as in:
https://github.com/KIT-MRT/mrt_cmake_modules/commit/105e18c370840c14ce1992fe88093bb01bf10a46

shr-project:
Backwards compatible fix for cv-bridge boost-python added in https://github.com/ros/meta-ros/pull/756

We might need similar patch for melodic recipes, but for those we should probably try to change them to respect ROS_PYTHON_VERSION as they are now searching for python38 (because only python3 support is enabled in boost recipe).

shr-project:
Converted to draft as this isn't strictly needed for hardknott anymore, so we can continue to use identical branch for dunfell/gatesgarth/master for a bit longer.

shr-project:
meta-ros-backports-gatesgarth: drop

is now merged in master and other changes were already implemented or fixed elsewhere, closing

PR #833, opened 2021-04-16

shr-project:


shr-project:
Merged in other branch as well:
   bab8a9da9..16bcd1820  build -> build

PR #837, opened 2021-04-21
[dunfell] spdlog work around to prevent prelink crashing in do_image

shr-project:


shr-project:
628796c95..4170b5d8c  dunfell -> dunfell-next

PR #840, opened 2021-05-01

shr-project:


shr-project:
Pushed to build-next as well:
   16bcd1820..33ad5b9b6  build -> build-next

PR #850, opened 2021-05-24
[build] Add support for ROS 2 Galactic Geochelone

shr-project:
https://discourse.ros.org/t/ros-2-galactic-geochelone-released/20559


shr-project:
Pushed to build-next as well:
   33ad5b9b6..fc860bda3  build -> build-next

PR #875, opened 2021-06-24
[build] Use */milestones/16 tags for meta-ros and meta-ros-webos

shr-project:


shr-project:
   fc860bda3..92fc415ed  build -> build-next

and tagged as:
   build/milestones/16

PR #880, opened 2021-07-02
[dunfell] meta-ros-common: add dependency on ros-backports-hardknott-layer

shr-project:


shr-project:
   8bcb179cd..ed0fce246  dunfell -> dunfell-next

PR #881, opened 2021-07-02
[gatesgarth] meta-ros-common: add dependency on ros-backports-hardknott-layer

shr-project:


shr-project:
   85d128c48..a2126a3e8  gatesgarth -> gatesgarth-next

PR #886, opened 2021-07-12
[build] Update to latest revisions of upstream layers and webOS OSE v2.12.0 build #366

shr-project:


shr-project:
  44ac9c26e..98f2dc19e  build -> build-next

PR #890, opened 2021-07-17
[build] Update to latest revisions of upstream layers

shr-project:

PR #891, opened 2021-07-23
[build] Indicate Support Level in the .mcf files

shr-project:
We're still updating some OE/ROS combinations even after they reach EOL. But we plan to do it only as long as it doesn't cause a lot of maintenance overhead.

But even the 'best-effort' support is different for e.g. zeus and gatesgarth build.
- meta-ros/zeus didn't receive any new features or ROS distribution updates since milestone/13 in 2020-07 and only couple build fixes were merged there. Last upstream layer update was merged shortly after in 2020-09 (and there were no changes in upstream since then).

- meta-ros/gatesgarth on the other hand is kept in sync with meta-ros/dunfell, so it receives all new features and ROS distributions updates like dunfell, hardknott, honister, just the upstream OE revisions aren't changing very often anymore (and .mcf files are still updated when needed).

It might be useful to somehow indicate this difference in the comment behind it.

Follow the policy from:
https://github.com/ros/meta-ros/wiki/Policies#openembedded-release-series-support
    
gatesgarth and zeus are now EOL according to:
https://wiki.yoctoproject.org/wiki/Releases
    
So we can drop zeus for LTS ROS distributions (foxy, dashing, melodic) from 'best-effort' to 'unsupported' (and from files/ to files-unsupported).

But there is no rush to remove gatesgarth builds as they are in sync with dunfell, I'm just going to build them less often especially for long "bitbake world" builds. But they are still useful when figuring out where some build issue was introduced between dunfell and hardknott.


shr-project:
Based on top of https://github.com/ros/meta-ros/pull/890, only last 3 commits are relevant for discussion here.

shr-project:
Gatesgarth was marked as EOL on 2021-06-08 in https://wiki.yoctoproject.org/wiki/index.php?title=Releases&diff=84746&oldid=84745

Zeus was marked as EOL on 2021-02-25 in https://wiki.yoctoproject.org/wiki/index.php?title=Releases&diff=84516&oldid=84514

shr-project:
   68b7a5dbd..199a6d4e7  build -> build-next

PR #894, opened 2021-07-29

shr-project:


shr-project:
This is the last update before overrides changes implemented in https://github.com/ros/meta-ros/pull/900

PR #895, opened 2021-07-30
[dunfell][gatesgarth][hardknott][honister] Convert to new override syntax

shr-project:


shr-project:
Replaced with a version for dunfell branch in https://github.com/ros/meta-ros/pull/902

PR #900, opened 2021-08-03
[build] Update to latest revisions of upstream layers with new overrides syntax

shr-project:
Depends on https://github.com/ros/meta-ros/pull/895, https://github.com/ros/meta-ros/pull/897, https://github.com/ros/meta-ros/pull/901 and https://github.com/ros/meta-ros-webos/pull/18

PR #903, opened 2021-08-05
[build] Update to latest revisions of upstream layers and webOS OSE v2.13.0 build #373

shr-project:
Depends on https://github.com/ros/meta-ros/pull/900 and https://github.com/ros/meta-ros-webos/pull/19


shr-project:
Pushed to other branches as well:
   ab87d9cf9..51caaf0eb  build -> build-next

PR #911, opened 2021-08-17
[honister] pendulum-control: blacklist, not compatible with glibc-2.34

shr-project:
* without an easy fix as reported in:
  https://github.com/ros2/demos/issues/530


shr-project:
   62bc4926b..068e939be  master -> master-next

PR #931, opened 2021-09-14
[honister][kirkstone] assimp: add from oe-core

shr-project:
* it was removed from oe-core in:
  https://git.openembedded.org/openembedded-core/commit/?id=92719e01a90f2a80f5cceb2f98a48c49f0c24f1c
  due to license issues, we should fix the license issues eventually as well
  e.g. by using debian tarball which removes the non-redistribute files
  https://lists.openembedded.org/g/openembedded-core/topic/85430468#155745
  https://lists.openembedded.org/g/openembedded-core/message/155757
  https://lists.openembedded.org/g/openembedded-core/message/155785

* there were some other issues discussed in:
  https://lists.openembedded.org/g/openembedded-core/message/155717
  https://lists.openembedded.org/g/openembedded-core/message/155219
  which broke qt3d from meta-qt6, but doesn't break ROS recipes

  in qt3d it got disabled in:
  https://code.qt.io/cgit/yocto/meta-qt6.git/commit/?id=6be91065b688fe691b691c13035e32b8a8dcda9c
  but webOS OSE is using older meta-qt6 revision, so I've blacklisted it in:
  https://github.com/shr-project/meta-webosose/commit/dd0b8891e541d5a05e52f35af881e104f9b19c65

* many ROS components depend on assimp and it was proposed for removal from
  oe-core also for hardknott in:
  https://lists.openembedded.org/g/openembedded-core/message/155752

Signed-off-by: Martin Jansa <martin.jansa@lge.com>

### DIFF
--- a/meta-ros-common/recipes-graphics/assimp/assimp/0001-assimp-remove-shared-lib-from-_IMPORT_CHECK_TARGETS.patch
+++ b/meta-ros-common/recipes-graphics/assimp/assimp/0001-assimp-remove-shared-lib-from-_IMPORT_CHECK_TARGETS.patch
@@ -1,0 +1,32 @@
+From: Ranjitsinh Rathod <ranjitsinh.rathod@kpit.com>
+Date: Fri, 13 Aug 2021 16:42:48 +0530
+Subject: [PATCH] assimp: remove shared lib from _IMPORT_CHECK_TARGETS
+
+In the target assimp::assimp shared library verification removed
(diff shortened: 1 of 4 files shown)


shr-project:
   bb96ccaf3..4e7ce48fc  honister -> honister
   bb96ccaf3..4e7ce48fc  honister -> honister-next
   bb96ccaf3..4e7ce48fc  master -> master-next

Issue #938, opened 2021-09-30
Packaging geographiclib datasets


windelbouwman:
I just faced the issue where mavros requires the presence of some geographiclib datasets. Those can be installed with this script:

https://github.com/mavlink/mavros/blob/ros2/mavros/scripts/install_geographiclib_datasets.sh

I'm willing to supply a patch as a bbappend to mavros to include these files, but is this the best place?

shr-project:
From the script it looks like we would need native version of ./meta-ros-common/recipes-extended/geographiclib/geographiclib_1.48.bb

and then it might make sense to add some PN-dataset package into geographiclib recipe and install the datasets there as the script as-is won't work in cross build environment anyway.

Then mavros could just have runtime dependency on geographiclib-dataset.

windelbouwman:
Yes, I agree, but I checked the scripts `geographiclib-get-*` and these are simple wrappers around wget.

The way I solved it now, is adding a `geographiclib-data` recipe, and added a `RDEPENDS` on `mavros`.

This is my `geographiclib-data` recipe. It is highly customized to mavros, so I'm in doubt whether it should be part of mavros instead, maybe the recipe should be named `mavros-geographiclib-datasets`?
```
# Mavros selects these files:
# geoids egm96-5
# gravity egm96
# magnetic emm2015

DESCRIPTION = "Data files for geographiclib."
SECTION = "devel"

SRC_URI = " \
    https://downloads.sourceforge.net/project/geographiclib/geoids-distrib/egm96-5.tar.bz2?use_mirror=autoselect;name=geoids;downloadfilename=egm96-5.tar.bz2 \
    https://downloads.sourceforge.net/project/geographiclib/gravity-distrib/egm96.tar.bz2?use_mirror=autoselect;name=gravity;downloadfilename=egm96.tar.bz2 \
    https://downloads.sourceforge.net/project/geographiclib/magnetic-distrib/emm2015.tar.bz2?use_mirror=autoselect;name=magnetic;downloadfilename=emm2015.tar.bz2 \
"

SRC_URI[geoids.md5sum] = "6ae1c02c0506a686c9ad33be42cad0ae"
SRC_URI[geoids.sha256sum] = "c46224f8f723dc915d97179f4e1580a98d6c742fe2b82cd8fef0ecaaad13e614"
SRC_URI[gravity.md5sum] = "dcc1ab8e3433bd0add901b2f3e5caa76"
[...]

PR #949, opened 2021-10-16
Add recipe geographiclib-data for mavros.

windelbouwman:
As discussed in #938 , this is a change proposal to add geographiclib data files specific for mavros.


shr-project:
Merged with https://github.com/ros/meta-ros/pull/951

PR #970, opened 2021-11-09

shr-project:


shr-project:
Merged to the other branch as well:
   310de57ff..140b1768e  build -> build-next

PR #975, opened 2021-11-19
[gatesgarth][hardknott][honister][kirkstone] {melodic} libsensors-monitor, cpu-temperature-diagnostics: remove bbappend

shr-project:
* lmsensors recipe was fixed with:
  https://lists.openembedded.org/g/openembedded-devel/message/93942
  we no longer need to remove the runtime dependency on it

* this can be also applied to dunfell once the fix is backported to meta-oe/dunfell as requested in:
  https://lists.openembedded.org/g/openembedded-devel/message/93942

PR #986, opened 2021-12-17
[dunfell] {melodic} libsensors-monitor, cpu-temperature-diagnostics: remove bba…

shr-project:
…ppend

* lmsensors recipe was fixed with:
  https://lists.openembedded.org/g/openembedded-devel/message/93942
  we no longer need to remove the runtime dependency on it

Signed-off-by: Martin Jansa <martin.jansa@lge.com>


shr-project:
Included in newer branches since https://github.com/ros/meta-ros/pull/975

   7399bf64c..d775fdbad  dunfell -> dunfell-next

PR #1008, opened 2022-05-02
Allow to build libg2o without x11 and opengl.

windelbouwman:
This probably applies to other yocto branches


robwoolley:
Merged into kirkstone-next

PR #1019, opened 2022-06-06
[build] Use */milestones/17 tags for meta-ros and meta-ros-webos

kuta42:


kuta42:
c49ba0822..fdc736c36  build -> build-next

and tagged as:
build/milestones/17

PR #1026, opened 2022-08-29
Superflore/noetic/2022 07 27

DasRoteSkelett:
Okay, this is my first shot at a superflore update to noetic. 

I can probably try to look for more rosdep stuff...


robwoolley:
Thanks, Mathias.  I have added support for Noetic in Kirkstone and Nanbield.  I haven't used ROS1 myself, so any feedback you can provide would be helpful.

PR #1051, opened 2023-09-28
Meta-ros Kirkstone

DRK-512:
Since kirkstone is an LTS release of Yocto, I thought it would be important to make a branch for it. 
I tested this with ros-core on kirkstone and everything seems to work. 
python3-empy used disutils which is not supported by python3 on kirkstone, so I switched it to setuptools3
there were also a number of licenses that were not accepted in this version of Yocto
- BSD just needed to be 0BSD
- meta-ros-common did not point at the licenses directory correctly
- cyclonedds contained an EDL license which was added to the licenses directory


DRK-512:
NOTE this PR will fail the 1000 file changed limit, that is due to every instance of LICENSE = BSD being changed to 0BSD to fit the kirkstone requirements

vmayoral:
Thanks for this @DRK-512. I think we want a tested LTS in mater branch indeed.

Anyone opposing to these changes? @robwoolley?

DRK-512:
@vmayoral if this does get approved and merged in, I will double check the humble stuff works on kirkstone. 
As of right now i only checked foxy and noetic and assumed it should be good. This was also pulled off the latest master branch so if humble does work on the master branch, there should be no issues here.

robwoolley:
@DRK-512 Thanks for the contribution.  I checked what 0BSD referred to.  It seems as if it is unrelated to the BSD license family, it is actually based on ISC: https://opensource.org/license/0bsd/

I suspect that these packages are actually under the 3-Clause BSD license.  I only checked asr_msgs, but confirmed that it is 3-Clause BSD not 0BSD: https://github.com/asr-ros/asr_msgs/blob/master/License.txt

I think the appropriate SPDX identifier is BSD-3-Clause (see https://spdx.org/licenses/BSD-3-Clause.html )

It seems like this change may have happened originally in Dunfell: https://lists.yoctoproject.org/g/yocto/message/49201

We'll want to make sure to propagate the fix to any other Yocto branches too.  Let me know if you need any help.

DRK-512:
Great catch @robwoolley 
I will change all the BSD0 into the BSD-3 clause format that's accepted by yocto then make another push. For now I will close this pr

Issue #1071, opened 2023-11-17
Add GitHub Actions to meta-ros


robwoolley:
Adding GitHub Actions to meta-ros would be useful for automating checks and builds.

- @ajlennon Yocto Layer CI Build and Test with GitHub Action by Alex Lennon (https://youtu.be/TsAcxd_acJI?si=HqRCqZ9M-f0T-Y-3) his work is here: https://github.com/DynamicDevices/meta-mono/blob/master/.github/workflows/CI_github.yml
- @razr did some work integrating meta-ros CI with GitHub Actions (https://github.com/razr/meta-ros/blob/hardknott-ur5-gh-action/.github/workflows/two-jobs.yaml)
- @Allegra42 is also giving a talk at the Yocto Summit 2023 entitled "Building a Yocto CI Pipeline with KAS, GitHub Actions and AWS - What could possibly go wrong?" (https://summit.yoctoproject.org/yocto-project-summit-2023-11/talk/RNHAPH/)

robwoolley:
@moto-timo Also pointed me to the work done for CROPS: https://github.com/crops/poky-container/blob/master/.github/workflows/build-test-deploy.yml

rcwoolley:
See [ROS 2 Tooling Working Group](https://github.com/ros-tooling/community) for existing ROS tooling:

- [Design: GitHub Action Release Process](https://github.com/ros-tooling/design/blob/main/docs/action-release.md)
- [action-ros-ci](https://github.com/ros-tooling/action-ros-ci)
- [ROS 2 GitHub Action CI Template](https://github.com/ros-tooling/action-ros-ci-template)
- https://github.com/ros-tooling/setup-ros
- https://github.com/ros-tooling/setup-ros-docker/tree/master
- https://github.com/osrf/docker_images

robwoolley:
See [ROS 2 Tooling Working Group](https://github.com/ros-tooling/community) for existing ROS tooling:

- [Design: GitHub Action Release Process](https://github.com/ros-tooling/design/blob/main/docs/action-release.md)
- [action-ros-ci](https://github.com/ros-tooling/action-ros-ci)
- [ROS 2 GitHub Action CI Template](https://github.com/ros-tooling/action-ros-ci-template)
- https://github.com/ros-tooling/setup-ros
- https://github.com/ros-tooling/setup-ros-docker/tree/master
- https://github.com/osrf/docker_images

robwoolley:
[...]

PR #1114, opened 2024-04-05
Superflore/rolling/2024 03 21 - Scarthgap

DasRoteSkelett:
This is somewhat recent ros2-rolling where ros-image-core will build with yocto scarthgap. The receipe-bbappends have been removed from the previous versions and build up from scratch. 

I guess there is still room for improvement to reduce the number of bbappends.

This should be merged into a scarthgap branch of meta-ros, I would recommend at least
@robwoolley


robwoolley:
Thank you, you have impeccable timing! I have syncs for all the supported ROS distros and Yocto releases queued up here and was about to merge them to the next branches after lunch: https://github.com/robwoolley/meta-ros/pulls

However, I still have some [failures for Scarthgap](https://github.com/robwoolley/meta-ros/pull/4#issuecomment-2039708311) that I was about to look into. It looks like you have already provided the fixes.  I hope to review the PR and merge them in ASAP.  (possibly this afternoon)

Regards,
Rob

README.md
robwoolley, line 1: Thanks for leading the charge in switching to kas.  I'm in favour of replacing mcf with kas as the recommended method to get started.  I had started with a very basic YAML file that builds Raspberry Pi 4: https://gitlab.com/ros-openembedded/oe-ros/-/blob/main/ros-kirkstone-humble-raspberrypi4-64.yml?ref_type=heads

kas-configs/limit-pressure.yml
robwoolley, line 1: I like this.  I do the builds on a shared build server and the builds are massive.  Having the bitbake pressure variables set by default is extremely helpful.

kas-configs/scarthgap-rolling.yml
robwoolley, line 1: This is excellent too.  I like the addition of diskmon.  My only suggestion is that to replace MCF we should use commit as well so we can provide kas YAML files for the milestones.

README.md
robwoolley: I just created a new issue #1115 to track these changes.  Would you mind breaking the kas changes into a separate PR?  I think it'll be easier to manage.

.github/workflows/github-action-build-scarthgap-rolling.yml
robwoolley, line 1: This is great to see as well.  In case you didn't see #1071 there are some helpful links to others that have used GitHub Actions for building OpenEmbedded.  Including @razr who was using it for ROS.

I had created GitHub Action workflows for superflore (https://github.com/ros-infrastructure/superflore/pull/301 ) and running the superflore tool for meta-ros (https://github.com/robwoolley/meta-ros/blob/kirkstone/.github/workflows/generate_recipes.yml )

For the builds themselves, I decided to use GitLab CI:  https://gitlab.com/ros-openembedded/poky-ros/-/blob/main/.gitlab-ci.yml?ref_type=heads

I think supporting both solutions is good.  I personally prefer using GitLab because I can store the repositories locally and avoid having to pull everything from across the network.  I also have been using the gitlab-runner's ability to run the builds from the command-line in a kind of "headless" mode that doesn't need to connect to gitlab.com.

In particular, I think GitHub Actions could be very useful for automatically generating SDKs and container images for ROS devcontainers.

Would you be open to creating a new PR for this work to keep it separate from the scarthgap changes?

robwoolley:
I reviewed all the commits and merged them into master-next: https://github.com/ros/meta-ros/commits/master-next/

Your fix for uncrustify was much better as it simplified things greatly.  I used the same approach of creating a new recipe for cpptoml to help simplify iceoryx-posh as well.

I found that I avoided some of the problems that were caused by CMake+Python by just upreving the rosidl-generator-py patch.  I still hit a problem with FindPython3, but this was fixed by following your recommendation of removing the older copy of the cmake recipe.  If there is anything I have overlooked with this approach that you were trying to solve, please let me know.

Lastly, I found that service-msgs had to be turned into a build dependency.  I was building for Raspberry Pi 4 and the msgs packages (like lifecycle-msgs and sensor-msgs) did not like it when I was trying to link the x86_64 library from service-msgs-native into the aarch64 binaries.

I like the other patches you included (removing unneeded patches, adding kas support, and GitHub Actions).   Since I can now build ros-image-core with just these commits, I would prefer to close this PR and review those under a new one.

Thanks again for this!  It came in right at the perfect time.

kas-configs/limit-pressure.yml
DasRoteSkelett: Unfortunately, your millage may vary a lot depending on which Server you have. At least to my knowledge, there is no one size that fits all. I added it here so I can append it to my jobs.

kas-configs/scarthgap-rolling.yml
DasRoteSkelett: I think this is totally valid for "releases". scarthgap has not even been officially released, therefore, there are quite some changes happening. I wanted to be able to always build the latest atm.

.github/workflows/github-action-build-scarthgap-rolling.yml
DasRoteSkelett: yes, it is much better to keep separate, also, I did not finish it. Thanks for the pointers.

DasRoteSkelett:
Hi @robwoolley ,
so please tell me, on which branch I should be working on. master-next? I can for sure also cross-compile to RPI.

robwoolley:
Hi @DasRoteSkelett 

I just merged master-next into master, so working on master should be fine now.  We are now fully up-to-date with the latest sync.  The image ros-image-core builds 100% and ros-image-world gets 99% of the way there with about 77 failures to investigate.  (ie 19858 tasks out of ~20000).

I think that covers everything in this PR.  I will go ahead and close it.  Please feel free to reopen it or create a new PR for anything I missed.

Regards,
Rob

PR #1117, opened 2024-04-10
Superflore/rolling/2024 04 08

DasRoteSkelett:
I cleaned up some of the bbappends that were not needed (at least as far as i can tell).


robwoolley:
This is difficult to review as there are many changes happening simultaneously.

Would it be possible to resubmit the PR with individual commits for each problem being solved so that they can be bisected in the future if needed?

Thanks,
Rob

DasRoteSkelett:
sure

PR #1119, opened 2024-04-11
{rolling} fix compile of rsl

DasRoteSkelett:
The include of rclcpp caused some configure time errors because some ament packages were not found. Unfortunately, these build time dependencies did not properly propagate, so we need to add them here alongside a patch to remove -Werror from compile flags (newer compiler always finds new warnings).

Also changed license of tl-expected to proper SPDX naming.


robwoolley:
Merged into master-next: https://github.com/ros/meta-ros/commit/70b2ce4fdb5eaeb386d367062c0f2b68b4513ae7

robwoolley:
Tested against Rolling for Styhead (ie master).  Merged into "master" branch.

PR #1120, opened 2024-04-11
{rolling} fix build of ecl-license

DasRoteSkelett:
This fixes two issues. The first one is that the package.xml is listening the wrong SPDX (BSD instead of BSD-3-Clause). The second one is that this package installs license files (into the wrong directory btw.) So to fix this properly, we only care for those files *iff* we build license packages. And if we do, we move them to the right directory.


robwoolley:
Merged into master-next: https://github.com/ros/meta-ros/commit/9b9c46cda9f1070a63f3d1e39970322ff10480e6

robwoolley:
Tested against Rolling for Styhead (ie master).  Merged into "master" branch.

PR #1121, opened 2024-04-12
{rolling} fix recipe append for shared-queues-vendor

DasRoteSkelett:
Updating the hash of the generated recipe. Unfortunately, a good solution (i.e. pulling the sources upstream & expect them to be in the same location) is a bit tricky.


robwoolley:
I am holding off on merging this as the 2 SRCREVs should be kept in sync.  For lack of an alternative, forcing both SRCREVs to be updated at the same time in the bbappend seems like the safest approach.

robwoolley:
Merged into master-next: https://github.com/ros/meta-ros/commit/976b0603a1ce83cb27d542bd87f1a97380a4ada7

PR #1126, opened 2024-04-15
{rolling} Update libphidget22-upstream to latest rolling

DasRoteSkelett:
Updated to version 1.19.20240304 which is currently used on rolling. Does no longer need the patch (which has disappeared upstream).


robwoolley:
Merged into master-next: https://github.com/ros/meta-ros/commit/33a55c86226a88d40e22080160d5ed3b17f60db4

robwoolley:
Tested against Rolling, Iron, and Humble for Styhead (ie master).  Merged into "master" branch.

PR #1193, opened 2024-07-03
{humble}: delete unneeded patch and appends

jiaxshi:
Missing cstdint include bug is fixed in rosbags project. https://github.com/ros2/rosbag2/pull/1321


jiaxshi:
The same patch is added in master/humble. I think the PR is applicable to master/humble.
For iron/jazzy/rolling the patch is not included.

robwoolley:
Merged into kirkstone: https://github.com/ros/meta-ros/commit/b4db6314bae1c78498158e267200a9c15a739cad

GustavoPimentel83:
@robwoolley, the patch https://github.com/ros/meta-ros/commit/b4db6314bae1c78498158e267200a9c15a739cad needs to be applied as well to the scarthgap branch to avoid compilation issues.

PR #1194, opened 2024-07-12
{humble} rosbag2-compression: remove outdated patch

stschake:
The missing include was added upstream in 0.15.11, so this patch is no longer needed (and in fact does not apply).


robwoolley:
Thanks, @stschake .  Merged into scarthgap-next.

PR #1240, opened 2024-10-22
pr2-msg: update SRC_URI

wico-frad:
The git repository pr2_common-release fetched by the recipe is no longer available, causing the build to fail during the fetch stage, so update the SRC_URI accordingly with the new link


robwoolley:
It looks like the git repository was restored.  Closing this PR and the related issue for now.  Please feel free to reopen it if needed.

PR #1254, opened 2024-11-10
{rolling,humble} zstd-vendor: Refresh integration patch

sgstreet:
The extends https://github.com/ros/meta-ros/commit/a0e9103651e3de6254c43580ebc6844b7f8e05ec to support ROS2 Rolling and Humble.


robwoolley:
I think this is already covered by 2e6401923da8edb11bd7ac6957c6811623944207.  I don't see a build failure in rolling for zstd-vendor.

PR #1257, opened 2024-11-13
Create a GitHub Workflow for OpenSSF Scorecard

robwoolley:
Adds scorecard.yml to automatically create a scorecard value using the OpenSSF Scorecard Action.


robwoolley:
Closing this without merging.  Unsure if there is any benefit to OpenSSF scorecards for Yocto layers.

PR #1277, opened 2025-01-02
Apply {common}{ros2} Add ROS opt prefix to PYTHONPATH to scarthgap

DasRoteSkelett:
This is a fix from @robwoolley missing in scathgap causing ament-cmake-core to fail

```
In order for Python to find modules in the recipe-sysroot we must add the path to PYTHONPATH.  However, we must prepend the path to any previous values of PYTHONPATH.

We cannot refer to the variable in its own definition since bitbake does not evaluate strings in that way.

Instead we must mimick python3targetconfig.bbclass and prepend to the recipe tasks to conditionally set PYTHONPATH in the shell context.
```


robwoolley:
Thanks, @DasRoteSkelett .  I have merged the change here: https://github.com/ros/meta-ros/commit/450a1afdc13b5696e43279f88e2286427ad2d3b2

PR #1278, opened 2025-01-06
{common}{ros2} (Backport) Add ROS opt prefix to PYTHONPATH Kirkstone

dirksavage88:
{common}{ros2} Add ROS opt prefix to PYTHONPATH

This is a backport based on fixes made by @robwoolley  on master. It allows users on kirkstone to build the ros2 interface packages that typically fail due to incorrect PYTHONPATH
```
In order for Python to find modules in the recipe-sysroot
we must add the path to PYTHONPATH.  However, we must prepend
the path to any previous values of PYTHONPATH.

We cannot refer to the variable in its own definition since
bitbake does not evaluate strings in that way.

Instead we must mimick python3targetconfig.bbclass and
prepend to the recipe tasks to conditionally set PYTHONPATH
in the shell context.
```


robwoolley:
Thanks, @dirksavage88.  I have merged it as https://github.com/ros/meta-ros/commit/1d8513f9cda8bbdf1d20062119baf474dfee52eb

PR #1290, opened 2025-01-28
{rolling} Add Qt fixes for styhead

robwoolley:
Fix Qt QA issues and add missing PYTHONPATH fix for styhead.


robwoolley:
Should be merged to other branches as well.

robwoolley:
Merged into rolling and styhead

PR #1299, opened 2025-02-06
{styhead} Make sdformat-urdf compile for ros2 control

sgstreet:
ROS2 Controls now depends on sdformat-urdf. For this to compile on rolling, update to Gazebo vendor packages is required, update:

* gz-cmake3 moves to gz-cmake4
* gz-utils2 move to gz-utils3
* gz-math7 moves to gz-math8
* sdformat moves to version 15


robwoolley:
Merged into multiple commits in master

PR #1302, opened 2025-02-07
{rolling} Update use-system-pykdl patch and add Upstream-Status

sgstreet:


robwoolley:
Merged into master: c70065e839c5349d3866eb5d3c24a90935963db9

PR #1313, opened 2025-02-23
canopen

vermaete:
When building 'canopen' on master (Yocto) and Jazzy (ROS) is needed these patches at Jazzy that where applied on Rolling.


robwoolley:
Thanks @vermaete .  I will take a look at merging this in.   I am glad you raised it as I have been looking to expand beyond desktop-full to include the meta-packages like canopen.

robwoolley:
I had to do some build system surgery to fix the buildpath issues, but I managed to get the canopen meta-package to build successfully on master-next.   I will port it to Jazzy soon.  Just wanted to let you know in case you were seeing the same problems that I was running into.

PR #1327, opened 2025-04-02
{noetic} actionlib: Fixed fuzz on patch

BasketsAlaaf:
On styhead, bitbake will not allow you to build with a patch that has fuzz. I simply updated the patch.


robwoolley:
Thanks, I will add it to my merge queue.

robwoolley:
I rebased this onto master-next because GitHub only seems to give credit for contributions that make it to master.  I will make sure to also merge this to styhead which is next in my list.

Thanks for the fix!

robwoolley:
Merged commit from master-next to styhead-next: https://github.com/ros/meta-ros/commit/1e6b4e9fd349d0283fc5238f11aa8706ed254f42

PR #1329, opened 2025-04-06
{noetic} Fixing infinite loop with roslogging and python3.12

BasketsAlaaf:
This fixes an issue described in
[this issue](https://github.com/ros/ros_comm/issues/2352#issuecomment-2287231681)

Basically, due to python3.12, which is default in styhead, the roslogging is stuck in an infinite loop trying to get the caller stack. Hence, roscore is at 100% CPU but does not work, nor does any other node.

This is a different solution as the suggested in the issue.


robwoolley:
Thanks, I will put this on the queue for merging in.

robwoolley:
I did a visual review.  This looks good.  I will add it to my merge queue.

PR #1330, opened 2025-04-06
{noetic} tf2: fixing compile with newer boost

BasketsAlaaf:
Seems like we are missing a header here, and the location of tuple type seems to have changed. This patch allows compiling of tf with styhead.


robwoolley:
I did a visual review.  This looks good.  I will add it to my merge queue.

PR #1367, opened 2025-04-23
{rolling} backward-ros: patch applied upstream

vermaete:
@see: https://github.com/pal-robotics/backward_ros/commit/423786d0155dc5372527ed7505116ed6ecfda865


vermaete:
And added the version bumps of the recipes I had to update.
+ fixed a few oelint-adv issues.

PR #1368, opened 2025-04-24
{rolling} fastdds: allow fastdds-native

vermaete:


robwoolley:
Thanks for reporting this.  There is another way to extend the build tools inside ros-distro.inc: https://github.com/ros/meta-ros/blob/master/meta-ros2-rolling/conf/ros-distro/include/rolling/ros-distro.inc#L23  It has the benefit of self-documenting what they are for in a convenient list.

I will cherry-pick the fix I put in for kilted instead: https://github.com/ros/meta-ros/commit/d79a59c1f91b093e1b41823a6220b77a53aef58b  I have merged it in here: https://github.com/ros/meta-ros/commit/1eb3038bda06bf795ad02140c36269c4ff6359bc

PR #1386, opened 2025-05-01
[scarthgap] {humble} ros2-controllers: resolve miscellaneous build failures and QA issues

graceagrace:
After the latest bump to version 2.44.0-1, several of the `ros2-controllers` recipes have build failures (missing dependencies) or QA issues (TMPDIR contamination). This PR resolves all such errors.

To test, build `ros2-controllers` while including the `meta-ros2-humble` layer in your `bblayers.conf` and confirm all controllers build successfully without QA issues.

### DIFF
--- a/meta-ros2-humble/recipes-bbappends/ros2-controllers/ackermann-steering-controller_2.49.1-1.bbappend
+++ b/meta-ros2-humble/recipes-bbappends/ros2-controllers/ackermann-steering-controller_2.49.1-1.bbappend
@@ -1,3 +1,9 @@
 ROS_BUILDTOOL_DEPENDS += " \
     generate-parameter-library-py-native \
 "
+
+# QA Issue: File /opt/ros/humble/share/ackermann_steering_controller/cmake/export_ackermann_steering_controllerExport.cmake in package ackermann-steering-controller-dev contains reference to TMPDIR [buildpaths]
+do_install:append() {
+    sed -i -e "s#${RECIPE_SYSROOT}[^;]*;##g" ${D}${ros_datadir}/${ROS_BPN}/cmake/export_${ROS_BPN}Export.cmake
+    sed -i -e 's#${RECIPE_SYSROOT}[^"]*##g' ${D}${ros_datadir}/${ROS_BPN}/cmake/export_${ROS_BPN}Export.cmake
+}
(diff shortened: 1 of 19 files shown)


robwoolley:
Thanks for the fix. I am eager to add ros2-controllers to my builds. I tried to build the packages with

> bitbake ackermann-steering-controller admittance-controller bicycle-steering-controller diff-drive-controller gpio-controllers gripper-controllers imu-sensor-broadcaster joint-trajectory-controller mecanum-drive-controller pid-controller pose-broadcaster range-sensor-broadcaster tricycle-controller tricycle-steering-controller

But I see failures control-toolbox and hardware-interface related to shadow declarations in those packages too.  I see that you removed the warning option pose-broadcaster.  Do you know why we are seeing the failure on OpenEmbedded but presumably it doesn't appear on Ubuntu?  I haven't had a chance to look into it.

robwoolley:
Thanks for the backward-ros fix.  I have merged that in to kirkstone, scarthgap, styhead, and master.

graceagrace:
Hi @robwoolley, thanks for the feedback - I dug into my build and I did make the appropriate changes to remove the `-Wshadow` error promotions for `control-toolbox` and `hardware-interface` in another project layer. I had forgotten about these!

As for the root cause, I plan to spend some time investigating this week. I do know that [this](https://github.com/ros-controls/ros2_control/pull/1181/files) PR and [this](https://github.com/ros-controls/ros2_controllers/pull/961/files) PR were where the changes were introduced - I am guessing we just recently saw these rolled out to Humble, hence what look like "new" build failures.

I'll take a deeper look (beyond just reverting the changes from these PRs) and see if it can be resolved without patches to every recipe - that'd certainly be ideal.

graceagrace:
Alright, I spent some time looking into this, and based on the error logs I am seeing for Humble (that have also been seen for Jazzy, as in https://github.com/ros/meta-ros/issues/1219), this is fundamentally a problem with `rclcpp`. Since the promotion of `-Wshadow` to an error was added to many packages under the `ros2_control` umbrella, these shadowed declarations in included headers are causing build errors. This is because of the global behavior of `-Wshadow`.

As for why this is not being seen on Ubuntu, my current theory is that it is simply due to differing versions of GCC. Further, `rclcpp` itself is not compiled with `-Wshadow=error`, so this build error would go undetected unless this error flag is applied in other packages including `rclcpp` headers, as with all the packages building as part of `ros2-controllers`, and you have a "problematic" compiler version. 

We have a few options for what to do next - I can add in my patches to `control-toolbox` and `hardware-interface` to remove `-Wshadow` and allow the `ros2-controllers` build to succeed. Another option would be to apply a patch to `rclcpp` to fix the shadowed declarations that the compiler is complaining about. This second option would ultimately result in fewer patches and would also fix compile errors for the `ros2_control` packages in Jazzy.

@robwoolley, let me know your thoughts, and I'll update my branch accordingly.

robwoolley:
@graceagrace  Thanks for the fixes.  Sorry it took me so long to get them merged.

PR #1400, opened 2025-05-22
[master] {kilted}{rolling} backward-ros

vermaete:
Hi Rob,

Two commits on backward-ros.  
On that was already at rolling and should also be at kilted.
An a small cleanup of the previous commit at rolling for backward-ros.


robwoolley:
Merged as d669caf11350cd198282772b274ebb6b0379ac0d for kilted
Merged as c6aec8b67d94e4f3fda37fa308a5e3f59b10e38f for rolling

PR #1427, opened 2025-06-22
[master] {rolling} Vermaete/master rolling 2025 06 09

vermaete:
hi @robwoolley 

My WIP on yocto master and your 2025 06 09 sync with rolling.

Do note that Yocto at the moment is bit unstable due to a change they make on the 'S = ' syntax.
This is by default not longer needed to have it in the recipe.
And the Yocto Project branched the Whinlatter releases.

For those reasons I'm using master-next for meta-openembedded and  openemedded-core right now.
Also some PR at meta-raspberrypi that are still open .


petpoly772:
hi

vermaete:
> hi

@petpoly772 ?

robwoolley (commented):
The yaml-cpp recipe we provide enables YAML_CPP_BUILD_CONTRIB and produces ${PN}-gtest.  I can't see any recipes that explicitly depend on it, but we may want to double check for conditionally enabled features before removing the functionality.

vermaete:
> The yaml-cpp recipe we provide enables YAML_CPP_BUILD_CONTRIB and produces ${PN}-gtest. I can't see any recipes that explicitly depend on it, but we may want to double check for conditionally enabled features before removing the functionality.

Hi Rob,
Just for clarity, these patches are on the previous sync of Master/Rolling.  I'm working now on the latest sync.
This was still WIP.  Not everything for desktop-full was compling.

I removed the yaml-cpp recipe because it is also in meta-openembedded/meta-oe.  For the this sync in the same version, but with some patches on top.  Now the recipe in meta-openembedded is a version newer. 
Well, the yaml-cpp recipe of meta-ros was anyhow not used in this setup because the one of meta-openembedded was taken.

robwoolley:
@vermaete I just completed getting master+rolling to build and pushed the changes to the master branch.  I tried to include all the fixes in your PRs.  Please let me know if there are any fixes that I missed.

robwoolley:
Did we catch everything from this PR?  Would it be okay to close it?

vermaete:
> Did we catch everything from this PR? Would it be okay to close it?

I will go over them now.  If not I will let you know which once I'm still missing.  If they needed a v2 version I will resend a PR.
And I will it when all of them are fine.

vermaete:
@robwoolley kind of half-way and only this one could be nice to have: https://github.com/ros/meta-ros/pull/1506

vermaete:
@robwoolley all done and working at [master]/{rolling} in current HEAD.

PR #1434, opened 2025-07-04
[scarthgap] launch-testing-ament-cmake: Set LICENSE

danascape:
* This fixes an issue while compiling scarthgap with AGL
* Backport the original patchset

Resolves: ERROR: launch-testing-ament-cmake-1.0.8-1-r0 do_create_spdx: Cannot find any text for license BSD

### DIFF
--- a/meta-ros2-humble/recipes-bbappends/launch/launch-testing-ament-cmake_1.0.10-1.bbappend
+++ b/meta-ros2-humble/recipes-bbappends/launch/launch-testing-ament-cmake_1.0.10-1.bbappend
@@ -1,0 +1,1 @@
+LICENSE = "Apache-2.0 & BSD-3-Clause"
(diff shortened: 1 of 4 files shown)


meta-ros2-jazzy/recipes-bbappends/launch/launch-testing-ament-cmake_3.4.3-1.bbappend
```diff
@@ -0,0 +1,3 @@
+# Copyright (c) 2025 Wind River Systems, Inc.
+
+LICENSE = "BSD-3-Clause"
```
nkoenig: [The original .bb file](https://github.com/ros/meta-ros/blob/scarthgap/meta-ros2-jazzy/generated-recipes/launch/launch-testing-ament-cmake_3.4.5-1.bb#L15C1-L16C1) has listed both Apache 2 and BSD. I think the issue is that the text `BSD` is no valid. So, should these changes be:
```
LICENSE = "Apache-2.0 & BSD-3-Clause"
```

PR #1450, opened 2025-07-06
kas: do check at master for the QA error 'license-exists'

vermaete:
If needed this ignore of this QA error should be set at recipe level.

What is already the case for openscenegraph:
  meta-ros2/recipes-support/openscenegraph/openscenegraph_3.6.5.bb:ERROR_QA:remove = "license-exists"


robwoolley:
Merged as 6476c605ff3a195353c0da5f50e80657f5bafe8b

PR #1456, opened 2025-07-07
[scarthgap] {jazzy} Use 3.x, not 4.x.y of cmake because gz_cmake_vendor-release uses 3.5.5

tomdeblauwe:
Use 3.x, not 4.x.y of cmake because gz_cmake_vendor-release uses 3.5.5. This makes gz-cmake-vendor compile.


robwoolley:
Thanks for catching this!

PR #1462, opened 2025-07-10
kas: ignore generated kas files and directories

vermaete:


robwoolley (commented):
This looks reasonable.  I personally don't do the builds in the same directly where I check out meta-ros build branch, but I see how this would be useful.

PR #1464, opened 2025-07-12
[master] {common} libyaml: fixed cmake 4.0.3 issue

vermaete:
Issue because Yocto/OE is not at cmake version 4.0.3

| CMake Error at CMakeLists.txt:2 (cmake_minimum_required):
|   Compatibility with CMake < 3.5 has been removed from CMake.
|
|   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
|   to tell CMake that the project requires at least <min> but has been updated
|   to work with policies introduced by <max> or earlier.
|
|   Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

### DIFF
--- a/meta-ros-common/recipes-support/libyaml/libyaml_0.2.5.bb
+++ b/meta-ros-common/recipes-support/libyaml/libyaml_0.2.5.bb
@@ -19,7 +19,8 @@ inherit cmake
 
 EXTRA_OECMAKE += "-DINSTALL_CMAKE_DIR=lib/cmake/yaml \
                   -DBUILD_TESTING=OFF \
-                  -DBUILD_SHARED_LIBS=ON"
+                  -DBUILD_SHARED_LIBS=ON \
+                  -DCMAKE_POLICY_VERSION_MINIMUM=3.5"
 
 DISABLE_STATIC:class-nativesdk = ""
 DISABLE_STATIC:class-native = ""


vermaete:
@robwoolley Could it be I'm missing this one for Master/Rolling and ros-image-core?

robwoolley:
CMake was upgraded in oe-core in the past week: https://git.openembedded.org/openembedded-core/commit/meta/recipes-devtools/cmake/cmake_4.0.3.bb?id=fc7aafb30bc5fe83f8d2ed451cb6b4d68b131fb5

Have you pulled oe-core recently?

vermaete:
Yep, a few minutes ago.  Well, living on the edge?

robwoolley:
Thanks, I understand what is going on now.  The latest CMake has a policy change that needs to be overridden to support older CMake versions.

I have updated the commit message to mimic the subject line Moritz was using on other recipes in meta-openembedded:

> {common} libyaml: Allow to build with CMake 4+
> 
> CMake 4.0.3 removes compatibility with CMake < 3.5 by default.
> 
> Set the policy for the minimum CMake version to allow it to configure
> anyway.
> 
> Signed-off-by: Jan Vermaete <jan.vermaete@gmail.com>
> Signed-off-by: Rob Woolley <rob.woolley@windriver.com>

PR #1466, opened 2025-07-12
[master] {common} cpptoml: fixed cmake 4.0.3 issue

vermaete:
| CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
|   Compatibility with CMake < 3.5 has been removed from CMake.
|
|   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
|   to tell CMake that the project requires at least <min> but has been updated
|   to work with policies introduced by <max> or earlier.
|
|   Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
|
|
| -- Configuring incomplete, errors occurred!


vermaete:
@robwoolley just as info
At master/rolling and ros-image-core I'm also needs this patch.

PR #1467, opened 2025-07-12
[master] {common} gz-math: added version 8.2.0

vermaete:
Needed for gz-math-vendor version 0.3.1 (@Rolling)

@robwoolley I left the older versions untouched.  I have no idea how to check if those are still needed for a ros/yocto combination.


robwoolley:
Thanks for submitting all the fixes.  I missed this and submitted a similar commit already to master branch.

PR #1472, opened 2025-07-12
{common} add ROS desktop image

vermaete:
Add a Yocto image that includes `ros-core` and `desktop-full` with X11 support.

@robwoolley I assume this commit was buried set of patches.  Could this one be added?

### DIFF
--- a/meta-ros-common/recipes-core/images/ros-image-desktop.bb
+++ b/meta-ros-common/recipes-core/images/ros-image-desktop.bb
@@ -1,0 +1,12 @@
+require ${COREBASE}/meta/recipes-graphics/images/core-image-x11.bb
+
+SUMMARY = "Desktop image of ROS 2"
+DESCRIPTION = "The desktop variant provides all commonly used libraries as well as visualization tools and tutorials."
+HOMEPAGE = "https://www.ros.org/reps/rep-2001.html"
+
+inherit ros_distro_${ROS_DISTRO}
+inherit ${ROS_DISTRO_TYPE}_image
+
+IMAGE_INSTALL:append = " \
+    desktop \
+"
--- a/meta-ros-common/recipes-core/images/ros-image-desktopfull.bb
+++ b/meta-ros-common/recipes-core/images/ros-image-desktopfull.bb
@@ -1,0 +1,12 @@
+require ${COREBASE}/meta/recipes-graphics/images/core-image-x11.bb
+
+SUMMARY = "Desktop Full image of ROS 2"
+DESCRIPTION = "The desktop_full variant provides a "batteries included" experience, enabling novice users to complete most entry tutorials without knowledge of the underlying library structure."
+HOMEPAGE = "https://www.ros.org/reps/rep-2001.html"
+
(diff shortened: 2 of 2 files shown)


robwoolley:
Coincidentally, I was in the middle of testing it.

Here is some feedback on the change:
* ros-core is already part of desktop-full so it is unnecessary to specify both (see https://www.ros.org/reps/rep-2001.html )
* we should probably just include desktop not desktop-full
* should we create a ros-image-desktop-full as well?  or just add perception and simulation to ros-image-desktop?
* It should automatically start into the desktop, right now the packages are included but I don't think anything is running on startup

vermaete:
Hi Rob,


> Coincidentally, I was in the middle of testing it.
> 
> Here is some feedback on the change:
> 
>     * ros-core is already part of desktop-full so it is unnecessary to specify both (see https://www.ros.org/reps/rep-2001.html )
> 
I can fix it.  But not anymore today.
>     * we should probably just include desktop not desktop-full
> 
It's for me the same.  I was just looking for something bigger than core to do a check on the recipes.
Having desktop and desktop-full seems to be perfect in line with rep-2001.

>     * should we create a ros-image-desktop-full as well?  or just add perception and simulation to ros-image-desktop?
> 
>     * It should automatically start into the desktop, right now the packages are included but I don't think anything is running on startup

For now, this seems to be an issue for later.  Yocto looks rather unstable on the master branch.  And meta-raspberrypi,the machine I'm using, is lagging behind.  But yes, I do agree that the desktop image should  be 'ready to be used'.

vermaete:
@robwoolley Thank for the link to the REP.  I made a v2 of the PR.  

BTW: nor desktop nor desktop-full is passing at the moment with MACHINE=raspberrypi5 and Yocto=master.
But I assume it could already be merged into master in the effort to get them passing.

robwoolley:
Thanks, I see having a desktop image recipe as something I would want to add across all combinations.  So, I don't feel the need to make master build 100% before merging it in.

I would like to better define what it is meant to be before merging it in.  For example, the ROS visualization tools currently require Qt5, X11, a graphics driver, and hardware acceleration.  I think that limits the use to Intel 64 (including qemux84-64) and Raspberry Pi 4 / 5 machines.  An added complication is that there are transitions going on at the moment from Qt5 with X11 to Qt6 on Wayland.   I suspect that Kilted and older will need to stay on Qt5 and Rolling should make the leap to Qt6.  I don't know what that means yet for weird combinations like Rolling on Kirkstone.

Some of these complications make me shy to merge in the image recipe before having a plan about where things are going.  It is an exciting problem to have though.  I see it as charting new waters for "desktop Yocto".

vermaete:
@robwoolley All true.  Another table in the readme.md file could help showing the users which image is working for which combination of Yocto/ROS at this moment.
But for now to have a big bunch of recipes passing in CI/CD it could be already useful.
I assume next on the list is 'world'.  But then we need the `python3-scipy` in the layers.

robwoolley:
Closing in favour of the new issue created in meta-oeros.

Issue #1477, opened 2025-07-13
python3-snappy: does not compile & why not from meta-pyhon


vermaete:
Hi Rob,

I have a question.  `python3-snappy` is a recipe in `meta-ros-common/recipes-imported-labgrid/python/python3-snappy_0.5.1.bb`.
And there is a recipe in `meta-python` at version 0.6.1.  Well, and it does compile at Yocto master.

Where could I find that this 0.5.1 is somehow required by a ROS package?
And if not, how can we be sure the recipe can be removed in meta-ros?

Br

BTW: the error

```
Log data follows:
| DEBUG: Executing shell function do_compile
| WARNING: '.' not a valid package name; please use only .-separated package names in setup.py
| /home/jan/projects/meta-ros/build/tmp/work/cortexa76-oe-linux/python3-snappy/0.5.1/recipe-sysroot-native/usr/lib/python3.13/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
| !!
| 
|         ********************************************************************************
|         Please consider removing the following classifiers in favor of a SPDX license expression:
| 
|         License :: OSI Approved :: BSD License
| 
|         See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
|         ********************************************************************************
| 
| !!
|   self._finalize_license_expression()
| In file included from /home/jan/projects/meta-ros/build/tmp/work/cortexa76-oe-linux/python3-snappy/0.5.1/recipe-sysroot/usr/include/python3.13/Python.h:72,
|                  from snappymodule.cc:28:
| snappymodule.cc: In function 'PyObject* maybe_resize(PyObject*, size_t, size_t)':
| /home/jan/projects/meta-ros/build/tmp/work/cortexa76-oe-linux/python3-snappy/0.5.1/recipe-sysroot/usr/include/python3.13/object.h:352:30: error: lvalue required as left operand of assignment
|   352 | #  define Py_SIZE(ob) Py_SIZE(_PyObject_CAST(ob))
|       |                       ~~~~~~~^~~~~~~~~~~~~~~~~~~~
| snappymodule.cc:71:13: note: in expansion of macro 'Py_SIZE'
|    71 |             Py_SIZE(str) = actual_size;
|       |             ^~~~~~~
[...]

PR #1483, opened 2025-07-21
[master] {common} labgrid cleanup

vermaete:
Removed two not longer compiling recipes in the labgrid directory.

@see: https://github.com/ros/meta-ros/issues/1477

I think none of these recipes in the labgrid directory are used.
And all of them are part of meta-python.
This PR is to stay on the safe side.  A bigger clean can be done later.


robwoolley:
Looks good.  One small detail is that these packages weren't added to support labgrid, they were imported from meta-labgrid which is why the recipes-imported-labgrid name was used to indicate where they came from.   I looked at the history and it seems like they were added to support Melodic and there may have been a version of rosbag that briefly used them for compression.  Likely to support the leveldb dependency which has since been dropped by rosbag.

vermaete:
This is better detective work.
I have seen those import directories a few time, but never thought about it.

PR #1494, opened 2025-07-25
[master]{rolling} rmw-zenoh-cpp: fixed 'find_if' is not a member of 'std';

vermaete:
added a patch that is also upstreamed to fix a gcc 15.1 issue


robwoolley:
I am not seeing this failure yet with rmw-zenoh-cpp_0.8.1-1.bbappend which was last changed by 676477418d337fb5a2692fb5f70980a752f21fe6

I will keep this PR open for when I see it.

robwoolley:
I couldn't reproduce the problem.  However, adding the missing include does not have any negative side effects.  I have merged the patch and tested that rmw-zenoh-cpp builds successfully with it.

PR #1495, opened 2025-07-28
[master]{common} gz-make4: added gz-cmake4_4.2.0

vermaete:
This version of gz-cmake4 is needed for gz-cmake-vendor-0.3.1

@see: https://github.com/gazebo-release/gz_cmake_vendor/blob/rolling/CMakeLists.txt#L6

Needed for ros-image-desktop


robwoolley:
Actually merged as 138cc48566b73472572199bfad8701a8d61ed735 in master

PR #1497, opened 2025-07-28
[master]{common} fcl: missing cassert include

vermaete:


robwoolley:
Thanks.  I am not seeing this yet with fcl-0.7.0 last changed by c9310c5ba3bf2575b13b1faf5935ca0d8b171cf2.

I will keep the commit open for when I run into it.

vermaete:
Strange — this patch no longer seems to be needed, even though the fix is not yet included in the current Rolling sync.

PR #1507, opened 2025-08-01
[master]{rolling} gripper-controllers: remove not longer used subdire…

vermaete:
…ctory

Clean up of c58fa256f17202ec03659823fb9c3d7159f935df. The recipe was removed but the directory with patches not.


robwoolley:
Merged as 38322f059599443a318a942db7fc945717e71839

PR #1513, opened 2025-08-04
{common} use openblas bbappend only with meta-python-ai

shr-project:
fixes:
https://github.com/ros/meta-ros/commit/efcda521f3dc9049d155b940221930e7f54e314d

meta-ros-common doesn't depend on meta-python-ai layer and now this bbappend was breaking parsing for builds without meta-python-ai.


robwoolley:
Thanks, @shr-project 

I have merged it into master.

PR #1539, opened 2025-08-21
babeltrace: drop bbappend

shr-project:
babeltrace recipe was dropped favour of babeltrace2 by: https://git.openembedded.org/openembedded-core/commit/?id=d16a5e05b3f383d21a1970c6613494416a7503c2

This was also causing 2 warnings at least since walnascar:
```
WARNING: oe-core/meta/recipes-kernel/lttng/babeltrace_1.5.11.bb: distutils-common-base.bbclass is deprecated, please use setuptools3-base.bbclass instead
WARNING: oe-core/meta/recipes-kernel/lttng/babeltrace_1.5.11.bb: distutils3-base.bbclass is deprecated, please use setuptools3-base.bbclass instead
```


robwoolley:
Thanks, Martin.  I heard Richard talking about this and agree with removing babeltrace 1.0 from OE.

Unfortunately, ros2_tracing still uses the babeltrace 1.0 APIs and the migration to 2.0 is still open: https://github.com/ros2/ros2_tracing/issues/22

I think the appropriate next step will be to duplicate the babeltrace recipe into meta-ros until we can confirm that this migration to 2.0 in ROS has been completed.

shr-project:
> Thanks, Martin. I heard Richard talking about this and agree with removing babeltrace 1.0 from OE.
> 
> Unfortunately, ros2_tracing still uses the babeltrace 1.0 APIs and the migration to 2.0 is still open: [ros2/ros2_tracing#22](https://github.com/ros2/ros2_tracing/issues/22)
> 
> I think the appropriate next step will be to duplicate the babeltrace recipe into meta-ros until we can confirm that this migration to 2.0 in ROS has been completed.

OK, please do that, I was just fixing the parsing error after oe-core update today, don't really use much from meta-ros anymore.

robwoolley:
I have made a copy of the babeltrace recipe from oe-core into all Yocto branches and merged the lines to enable the Python binding.

PR #1549, opened 2025-08-24
kas: add the Kconfig file for using kas `menu kas/Kconfig`

vermaete:
@robwoolley My latest version of 'kas menu'.

I would keep it a while in the repo but not yet as a replacement of the toplevel kas files.

br,


vermaete:
I'm closing my own PR.  The bitbake-setup looks like a good replacement of Kas.

PR #1569, opened 2025-09-21
{rolling}: oelint: Assignment should be 'VAR += "foo"' not 'VAR += " …

vermaete:
…foo"'

fixing the oelint-adv issues as:
  error:oelint.vars.inconspaces:Assignment should be 'VAR += "foo"' not 'VAR += " foo"'

@robwoolley I did only one rull of oelint in meta-ros2-rolling and only the bbappends.
I think there also two PR's open in superflore to fix some low risk oelint warnings in the generated bb files


robwoolley:
Thanks @vermaete . This is great work to have.  This is very timely as the monthly syncs are coming in that change the generated recipes and I can hopefully test both sets of changes at the same time.

PR #1570, opened 2025-09-22
{rolling} ros-distro: delete not longer valid ROS_UNRESOLVED_DEP variables

vermaete:
Removed not longer used ROS_UNRESOLVED_DEP variables.
@robwoolley Could you please review.  It's still a rather dark corner of meta-ros/superflore for me.

And many more to come :-)


robwoolley:
Thanks, I'll check this out.

PR #1575, opened 2025-09-27
[scarthgap]{jazzy}: rmw-zenoh: use system zenoh (zenohc/zenohcxx)

schorsche32:
Make rmw-zenoh use system-provided zenoh libraries and adjust dependencies. This mirrors the approach in meta-ros2-rolling, with the addition of fastrtps-cmake-module-native which appears required here.


robwoolley:
Thanks for pointing this out @schorsche32.

I am in progress of merging it in.  As a side note, I discovered a mistake in my original patch where I think the second SATISFIED line should be SATISFIED zenohcxx_FOUND.

robwoolley:
Quick update: merged in master as 052d761feb0cd9537fb72fdc28294fe6566b016f

Scarthgap forthcoming

schorsche32:
@robwoolley Thanks!

PR #1576, opened 2025-09-30
scripts: add the script and the arguments into the generated git comm…

vermaete:
…it message

This will clarify how the git commit was generated.

@robwoolley It could not run the scripts.  So I could not test the changes.  "But it should work" TM

Having multiple -m or --message to a git commit is allowed and will put the messages in new paragraphs


vermaete:
@robwoolley Would you like to consider merging this one in?  Our do you want some changes?

robwoolley:
@vermaete It looks fine as-is.  Thanks for raising it to my attention.  I will give it a test run and look to merge it.

PR #1582, opened 2025-10-03
{rolling} removed duplicated LICENSE variable in bbappends

vermaete:
Removed the LICENSE variable if the value in bbappend is the same in the generated bb file.


vermaete:
This one does not apply anymore at rolling/master.
I will send a new one for the latest sync of rolling/master.

PR #1583, opened 2025-10-05
{rolling}: replace ${PYTHON_PN} by python3

vermaete:
see: https://docs.yoctoproject.org/singleindex.html#removed-variables

There are still some ${PYTHON_PN} strings in the generated bb files. This should be solved in rosdistro.

tested at rolling/master


vermaete:
This one does not apply anymore at the upcoming sync of master/rolling.
I will send a new one.

PR #1593, opened 2025-11-05
[whinlatter]{common} Remove S = "${WORKDIR}/git" from recipes

johannesschrimpf:
Yocto Whinlatter throws an error if a recipe sets S = "${WORKDIR}/git" since this is already set by bitbake.conf in oe-core:

```
ERROR: stress-1.0.7-r0 do_unpack: Recipes that set S = "${WORKDIR}/git" or S = "${UNPACKDIR}/git" should remove that assignment, as S set by bitbake.conf in oe-core now works.
```


robwoolley:
Thanks for contributing this, @johannesschrimpf.

I removed the offending line in this commit which will be present in whinlatter when I create the branch: 5239278a0f9ff6857cf9d4fe280eb0cec66943b4

Sadly, I made the change myself as well instead of merging your PR.  Sorry about that.  I try to merge changes from others first but was behind on the syncs and trying to get back on top of it.  I look forward to seeing you contribute again in the future.

Issue #1615, opened 2025-11-24
Merge support for mash


robwoolley:
The mash tool (https://github.com/[robwoolley/mash](https://github.com/robwoolley/mash) inherits mash_generated.bbclass

Integrate https://github.com/robwoolley/meta-ros/blob/spaceros-preview/meta-ros-common/classes/ros_generate_generated.bbclass as mash_generated.bbclass replacing ros-generate with mash-generate, ROS_GENERATE_GENERATED, with MASH_GENERATED, and ROS_GENERATE_GENERATED_BUILDTOOLS with MASH_GENERATED_BUILDTOOLS.

robwoolley:
This is the commit with the old tool name: https://github.com/ros/meta-ros/commit/1b69396b1b317243b152f58b0260398296c6dee6

PR #1618, opened 2025-11-26
{common} Add mash_generated.bbclass

robwoolley:
Add a mash_generated.bbclass to support recipes built using the mash tool (https://github.com/robwoolley/mash)

Resolves issue: https://github.com/ros/meta-ros/issues/1615


robwoolley:
The mash_generated.bbclass is present on master, whinlatter, and scarthgap.  It is absent from kirkstone, however kirkstone will be EOL soon and mash hasn't been tested for kirkstone.

PR #1629, opened 2025-12-21
Vermaete/meta virtualization

vermaete:
A new dependency of meta-python-ai


robwoolley:
Thanks @vermaete 

I want to start experimenting with containers at runtime which also requires the meta-virtualization layer.  I added it as its own layer YAML file and included it to common.yaml:

https://github.com/ros/meta-ros/commit/a3d0a00023b69050db789db206795074ef57d2a0
https://github.com/ros/meta-ros/commit/614ec3e12befc41f1cd48a117e2e81f7c5074134

PR #1630, opened 2025-12-21
Vermaete/master rolling 2025 12 04

vermaete:


vermaete:
@robwoolley This set of PR should bring ros-image-desktop working on qemu and RPi5.  *should* as: I'm still compiling. I'll leave ros-image-desktopfull for what it is.  And put some effort on the ptests.

vermaete:
@robwoolley last two commits are the current status of the ptest.  I would not yet add them to master branch.

I have tested it at a Raspberry Pi 5 on the recipe `sensor-msgs-py`.
You have to add `inherit ros-ptest-python-pytest` to the bbappend file.
And run them with `ptest-runner sensor-msgs-py`.
    
And they passed :-)

See you in the next meeting.

meta-ros-common/recipes-core/images/ros-image-desktop.bb
robwoolley, line 1: Hi @vermaete.  Thanks for keeping this going.  I have been thinking about what the right thing to do is for the desktop image recipe.   I would like to define what it is independently from the desktop and desktop-full image variants.   I feel that ros-image-desktop should have everything that a user may expect.  So I'm leaning to making desktop-full the dependency for it.  I would also expect it to automatically boot right into the desktop.  For rolling it should also be using Qt6 with hardware acceleration to ensure that Gazebo and RViz are running optimally.

I'm going to skip this commit in this series, but I'd like to keep working on this and help you out with it however I can.

meta-ros-common/recipes-devtools/python/python3-pydocstyle_6.3.0.bb
robwoolley, line 1: Unfortunately, I can't take thevpython3-pydocstyle recipe from meta-sca.

The maintainer of meta-sca has previously contacted us asking that we not duplicate his recipes.  I want to respect his wishes. My recommendation would be to create a new recipe using recipetool.

As an FYI, the recipes in meta-sca are also under the BSD-2-Clause instead of the typical MIT license.  So we would need to add a copyright as a minimum:  https://github.com/priv-kweihmann/meta-sca/blob/master/LICENSE

robwoolley:
@vermaete:  Thanks again for all the hard work put into cleaning up the recipes.  I really appreciate it.   I went through each commit and cherry-picked them into my local branch for master (which I will be pushing soon).  There were only a few commits that I had to skip for now.  Notably the python3-pydocstyle recipe, the ros-image-desktop-* image recipes, and the last 2 commits marked WIP about ament-xmllint and ros-ptest-python-pytest.

If it works for you, I suggest we close this PR and create new, individual PRs to track the outstanding work.

vermaete:
done

PR #1654, opened 2026-02-02
Update pcl-ros_2.6.2-1.bbappend for TMPDIR

tomdeblauwe:
Other packages have a better sed to fix the local paths in TMPDIR problem. This change adds the same regex here.


robwoolley (changes requested):
Thanks @tomdeblauwe 

I infer that the new line 8 specifically handles the case where RECIPE_SYSROOT is the last path in the list.

Should the new line 8 be changed to add an additional double quote after the asterisk to avoid it being run in unexpected cases?
```
    sed -i -e 's#${RECIPE_SYSROOT}[^"]*"##g' ${D}${ros_datadir}/${ROS_BPN}/cmake/export_${ROS_BPN}Export.cmake
```

tomdeblauwe:
yes, correct. it is replacing the sysroot path + all non-quote characters, so it will stop at a quote.

edit: ok, true: by adding the quote you make it indeed more robust

Issue #1653, opened 2026-02-02
[scarthgap][walnascar] Remove python3-pydantic on older releases


robwoolley:
Backport fceba46917fe9473e64534d972d3cf853aa7674d to remove python3-pydantic on walnascar and scarthgap.

python3-pydantic is not present in meta-python in kirkstone.  Leave the recipe there.

robwoolley:
Fixes were merged into scarthgap and walnascar.

PR #1657, opened 2026-02-03
[walnascar] {ros2} python3-pydantic: Drop recipe

shr-project:
There is 2.12.0 version in meta-python, this 2.5.3 is too old to build python3-lief.

Fixes https://github.com/ros/meta-ros/issues/1653


robwoolley:
Thanks @shr-project, I will merge this in ASAP.

PR #1660, opened 2026-02-05
{kilted} apriltag: Fix build due to compiler warning

eigendude:
## Description

While trying to build the apriltag package for my Yocto OS (kilted/scarthgap) I get a compiler warning in a numpy header, breaking the build.

I fixed this with `OECMAKE_C_FLAGS:append = " -Wno-error=pedantic"`.

My Yocto OS uses scarthgap/kilted, so I applied it there for testing, but can retarget different versions if the patch is wanted.

## How has this been tested?

Before, build fails with error:

```
| apriltag/3.4.3-2/recipe-sysroot-native/usr/lib/python3.12/site-packages/numpy/core/include/numpy/__multiarray_api.h:847:11: error: ISO C forbids conversion of object pointer to function pointer type [-Werror=pedantic]
|   847 |         (*(PyObject * (*)(PyTypeObject *, int, npy_intp const *, int, npy_intp const *, void *, int, int, PyObject *)) \
|       |           ^
| apriltag/3.4.3-2/recipe-sysroot-native/usr/lib/python3.12/site-packages/numpy/core/include/numpy/ndarrayobject.h:118:9: note: in expansion of macro 'PyArray_New'
|   118 |         PyArray_New(&PyArray_Type, nd, dims, typenum, NULL, NULL, 0, 0, NULL)
|       |         ^~~~~~~~~~~
| apriltag/3.4.3-2/git/apriltag_pywrap.c:269:42: note: in expansion of macro 'PyArray_SimpleNew'
|   269 |         xy_lb_rb_rt_lt = (PyArrayObject*)PyArray_SimpleNew(2, ((npy_intp[]){4,2}), NPY_FLOAT64);
|       |                                          ^~~~~~~~~~~~~~~~~
| cc1: all warnings being treated as errors
```

After applying this patch (and also https://github.com/AprilRobotics/apriltag/pull/424), build succeeds.

PR #1661, opened 2026-02-06
{all} Fix apriltag/apriltag-ros

eigendude:
## Description

This PR replaces https://github.com/ros/meta-ros/pull/1660. It expands the scope by fixing apriltags and apriltags-ros for all ROS distros.

## How has this been tested?

Both patches tested on my Scarthgap/Kilted build.

Currently working on porting my whole OS to master/rolling (in preparation for wrynose LTS) but haven't gotten it to build yet.

### DIFF
--- a/meta-ros1-noetic/recipes-bbappends/apriltag-ros/apriltag-ros/0001-update-tf2-convert-header.patch
+++ b/meta-ros1-noetic/recipes-bbappends/apriltag-ros/apriltag-ros/0001-update-tf2-convert-header.patch
@@ -1,0 +1,43 @@
+Upstream-Status: Backport [https://github.com/christianrauch/apriltag_ros/commit/5b73d4fd3eccdd0ccc6d17043f9f401ca04f6a40]
+
+Signed-off-by: Garrett Brown <garrett.brown@aclima.earth>
+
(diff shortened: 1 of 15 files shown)


robwoolley:
Thanks @eigendude.   I appreciate you separating the two fixes and providing individual commits for each branch.

May I ask you to replace the sed line that fixes the deprecated tf2 headers with a patch instead?  I try to only use the sed statements on generated files (eg. in the do_configure() or do_install() task).  It is preferrable to use a patch when modifying the source code.

If there is an upstream commit that fixes it that is even better as it gives me a clue to drop the patch in a future release.  In that case you will need to add Upstream-Status: with Backport or Pending as well as your Signed-off by: line.

eigendude:
I'll see if there's a commit to backport. sed isn't ideal because it works different on macOS. But it does noop when upstream gets a new release.

eigendude:
Updated with patch from upstream.

EDIT: Oops, missing upstream-status. Will update. What should I set it to, given that the patches are _from_ upstream?

robwoolley:
Hi @eigendude,  I started to merge your PR yesterday.  I am doing it on master first, so I ran into some package QA issues with camera_ros and apriltag_ros.  Those are fixed now and I plan to push shortly.

PR #1675, opened 2026-03-25
Move opencv bbappend to meta-ros1-noetic

apbr:
It seems this bbappend is only required for meta-ros1-noetic.

Unfortunately I couldn't test it across all ROS versions. Is there a CI for that?


robwoolley:
Thanks @apbr.  Sorry for the delay in responding.  It looks like a good catch.  I will cover testing it.  I haven't hooked up an automated CI to GitHub.

apbr:
Thanks

PR #1686, opened 2026-04-10
Work-in-Progress: Qt6 Support

robwoolley:
This draft PR enables building the ROS 2 variants: desktop and perception with Qt6

It is intended as a preview and points directly to unmerged PRs in ros-visualization that add Qt6 support.


robwoolley:
This should build these targets successfully:

bitbake --continue desktop perception ros2-control canopen

robwoolley:
This has been replaced by a newer series #1711

PR #1698, opened 2026-04-15
{common} python3-pyflakes: remove recipe

vermaete:
The cecipe at meta-python is more recent.

@see: https://github.com/ros/rosdistro/pull/50075


robwoolley:
This has been added as 5d2ec9271dc35deba05c53444646cfee765887a6 in master-next.  It will be merged into master shortly.

PR #1712, opened 2026-05-03
{rolling}[master] few patches

vermaete:
@robwoolley please review.  If possible it would be great to have them at master before the next sync of rolling


robwoolley:
Sounds good, I will include them on the current Rolling sync that I am working on.

This sync brings in Qt6, so there will be a lot of changes to the recipes for Gazebo and the visualization tools.

robwoolley:
@vermaete Thanks, I merged the commits.  I confirmed all recipes build successfully except fields2cover which had a dependency ortools-vendor which wouldn't build due to a minimum CMake problem.  It was also compounded by it trying to fetch dependencies during the compile task.  They probably should be fetched by bitbake in advance.

PR #1720, opened 2026-05-08
[master]{common} boost: Stop requiring NumPy for Boost.Python

eigendude:
## Description

This PR removes the `python3-numpy-native` dependency from boost. It should instead be added by downstream consumers of boost.

Boost.Python can be built without NumPy. The meta-ros Boost append adds `python3-numpy-native` unconditionally and packages `libboost_numpy` with the Boost.Python package, causing unrelated Boost users to depend on NumPy.

Only package Boost.Python here. Recipes that require Boost.NumPy should enable and depend on it explicitly instead of making all Boost users depend on NumPy.

## Motivation and context

It was increasing the time of my build without ROS, from 63mins 50s, to 64mins, 32s.

## How has this been tested?

Tested on my work OS, based on Scarthgap, with this change backported.

### Before

* Build tasks: 7293
* Build time: 64mins 32s

### After

* Build tasks: 7279
* Build time: 63mins 50s

### Additional verification: Remove meta-ros entirely

* Tasks: 7278
* Build time: 63min 52s

(The final task I traced down to a summary or benign action or something.)

### Additional verification: Blocklist `python3-numpy-native`

To scrutinize the size of my OTA, I did:

```bitbake
SKIP_RECIPE[python3-numpy-native] = "${@'python3-numpy-native is not allowed.' if d.getVar('BUILD_DEVELOPMENT') == '0' else ''}"
```

Before:

My build failed because boost pulled in `python3-numpy-native`.

After:

Build succeeded because numpy-native wasn't built.

### DIFF
--- a/meta-ros-common/recipes-support/boost/boost_%.bbappend
+++ b/meta-ros-common/recipes-support/boost/boost_%.bbappend
@@ -2,11 +2,8 @@
 
 inherit python3native
 
-DEPENDS += "python3-numpy-native"
-
 FILES:${PN}-python = " \
     ${libdir}/libboost_python*.so.* \
-    ${libdir}/libboost_numpy*.so.* \
 "
 
 do_configure() {


jpao79:
What was causing the gFortran requirement?  

I couldn't figure out why it was already checking for gFortran installation when I was just building curie core which doesn't have any ROS requirements that I could tell.

jpao79:
Briefly looking at it, I think what maybe happening from my limited knowledge is that something we do need actually needs boost and the meta-ros layer is modding the boost recipe and was adding numpy.

So seems like this is one way to solve it, but somewhat unnecessarily complicated since we actually don't need meta-ros layer.  

The cleanest way to fix the build would be to remove the meta-ros layer (which we aren't currently using at the moment) and if we do need to use it in the future we can add this fix (or add gFortran as a req).

jpao79 (approved):
👍 if we really want to keep meta-ros for some reason.

eigendude:
@robwoolley any appetite for this change?

robwoolley:
@eigendude Thanks for submitting this.  I am open to merging this as I think removing the unnecessary dependency will be very helpful in reducing size of the runtime, build, and SDK size.  I hope to test it soon after I get the latest releases merged in.

eigendude:
Cool, take your time, I'm just carrying the patch in our local `scarthgap` branch.

robwoolley:
Quick update:  I tracked this down further and submitted a fix upstream to the b2 build system used by boost: https://github.com/bfgroup/b2/pull/611   I am waiting to hear back what they think.

Meanwhile, I have submitted it to oe-core and it has been merged into master:
https://git.openembedded.org/openembedded-core/commit/?id=4e52a93832b634ac36ea9e11582d92214c89150b
https://git.openembedded.org/openembedded-core/commit/?id=7b41a386237a4b7d8b4b750740d895d6e524c74d
https://git.openembedded.org/openembedded-core/commit/?id=975ecfc285adfb4af7f53032ffbccbbac6a216ed

Let's keep this issue open for now for me to track backporting the fix to Scarthgap and Wrynose.

eigendude:
Nice! This issue really rabbit-holed there.

PR #1744, opened 2026-06-01
[master]{common}{rolling} Add ptest infrastructure

vermaete:
@robwoolley some work on the meta-common/recipes-intrastructure

* Bringing them to the latest version
* remove the not longer needed .inc files (from those python2 days)
* adding ptest for those recipes where the tests are distributed (not always the case with pypi)

Not all ptests are passing.  But adding the ptest code to the recipe isn't breaking anything.

I didn't test the new versions on the target.  I don't even have a clue how to, not how to be sure the coverage is okay.
But that's for most of the recipes the case, no?

I would keep it in draft until more tests are passing.


robwoolley:
@vermaete I took a look at merging this.  I kept running into this error:

```
ERROR: Nothing PROVIDES 'meta-ros-image-ptest'
meta-ros-image-ptest was skipped: No class extension set
```

I added the following to my conf/local.conf:
```
DISTRO_FEATURES:append = " ptest"
```

Is there anything else that I should do to test this out?

vermaete:
Hi @robwoolley 

I will split this PR into smaller pieces.  One for every recipe + one for the 'global' files.
And add my script how to run the ptests into the commit of the 'global' files or into meta-oeros.

br

PR #1755, opened 2026-06-24
[wrynose] Remove invalid LICENSE for off-highway-can bbappends

grsandeep85:
Remove invalid LICENSE for off-highway-can bbappends in rolling, jazzy and humble.

The bbappend set LICENSE = "Apache-2.0 && MIT" using '&&' which is invalid BitBake syntax and causes a parse error:
      error in 'Apache-2.0 && MIT': invalid syntax
    
The correct operator is '&'.  Remove the bbappend entirely so the base generated recipe's LICENSE value is used without the invalid override.


grsandeep85:
@robwoolley can you prioritize merging this patch.

robwoolley:
Thanks @grsandeep86, I will prioritize merging the patch.

PR #1776, opened 2026-07-12
[master]{rolling} ros-distro.inc: lz4 is resolved

vermaete:
* defined in rosdep of rosdistro
* lz4 recipe at openembedded-core


vermaete:
@mergifyio queue

PR #1778, opened 2026-07-13
[wrynose]{lyrical}{ros2} generate-parameter-library, tl-expected: Remove obsolete patch, allow empty package

maxweberhohengrund:
Two related fixes for building generate-parameter-library 1.2.0 on lyrical:

1. Remove the obsolete use-upstream-tl-expected.patch bbappend for lyrical.
   The changes from this patch were merged upstream in
   generate-parameter-library 1.2.0 (the backward-compatibility tl_expected
   references were dropped), so the patch now fails to apply and do_patch
   errors out. The rolling (0.8.0) and jazzy (0.7.1) bbappends still apply
   to their versions and are left untouched.

2. Set ALLOW_EMPTY:${PN} = "1" in the shared tl-expected recipe. tl-expected
   is a header-only library, so its main package is empty and is normally
   not generated. However, generate-parameter-library depends on it at
   runtime via the ROS_UNRESOLVED_DEP-libexpected-dev mapping, so image
   dependency resolution fails unless the empty package exists. This follows
   the same approach as the existing libeigen bbappend in meta-ros-common.

Build-tested on wrynose/lyrical: do_patch of generate-parameter-library
succeeds and images that pull in generate-parameter-library assemble
correctly with the empty tl-expected package.


robwoolley:
Thanks @maxweberhohengrund for pointing this out.

The fix to generate-parameter-library was part of the massive 400+ commit that I just committed to wrynose: https://github.com/ros/meta-ros/commit/f7eabb56b5e134709217da03386bb5f65dd5ee4e

Regarding the tl-expected recipe, that is a really good catch.  My preference would be to backport the fix from the other branches to keep the context the same: https://github.com/ros/meta-ros/commit/949fd1e98eb17f5f1b7ea3b60bcf15e7da77e413

I'll use this PR to track that I need to do that.

robwoolley:
Thanks for reporting this.  I backported the fix to keep the branches the same: 459d028311a61dbca9f148b8d59cf114fa2ee210

Closing as fixed.

PR #1781, opened 2026-07-13
[wrynose]{lyrical} foonathan-memory-vendor: Fix system foonathan-memory detection

maxweberhohengrund:
Building foonathan-memory-vendor 1.4.1-1 on lyrical fails: do_configure
reports "foonathan_memory not found." although foonathan-memory is staged
in the sysroot (via this layer's existing bbappend), and the vendor then
falls back to cloning github.com/foonathan/memory at build time, which the
network-isolated do_compile blocks ("Could not resolve host").

Since 1.4.1 the vendor checks for the system library by spawning a child
cmake via `cmake --find-package` (deprecated CMake functionality), which
does not inherit the cross toolchain file, CMAKE_SYSROOT or
CMAKE_FIND_ROOT_PATH. Kilted's vendor 1.3.1 checked in-process and was
unaffected. The patch switches the check to a normal in-process
find_package(), which respects the toolchain context.

This blocks the default Fast DDS path on wrynose/lyrical. Verified in
production image builds on wrynose/lyrical. We intend to submit the fix to
eProsima/foonathan_memory_vendor as well and will update the
Upstream-Status accordingly.


robwoolley:
Thanks @maxweberhohengrund .  This seems like an important fix.  I will try it out and add it to my merge queue.

robwoolley:
I wasn't able to rebase the PR branch, so I cherry picked the commit and merged it here: 2ebdb0b8691a3611015cb49afc4ef2959dd599a6

Thanks again for this fix, @maxweberhohengrund !  Very appreciated.

PR #1784, opened 2026-07-21
Vermaete/space ros

vermaete:
Two typo's in the value of the generated LICENSE variable of the bitbake recipe.
For both a change was send to the upstream projects.
I assume no need for a bbappend.  Both could be applied before a new sync.


vermaete:
sorry, was already fixed in the bbappend file

PR #1796, opened 2026-08-08
[master]{common} python3-rospkg: version bump 1.3.0 -> 1.6.1 + ptest

vermaete:
* removed not longer needed .inc file
* added ptest

* note the patch on `os_detect.py`.

  The 'os' 'openembedded' is now discovered by the lsb info.
  But can still be overrulled by the ROS_OS_OVERRIDE variable.
  
  e.g:
  
  ```
  root@qemux86-64:/usr/lib/python3.14/site-packages# python3 -m rospkg.os_detect
  OS Name:     openembedded
  OS Version:  nodistro.0
  OS Codename: 
  
  root@qemux86-64:/usr/lib/python3.14/site-packages# export ROS_OS_OVERRIDE=ubuntu:18.04:bionic
  
  root@qemux86-64:/usr/lib/python3.14/site-packages# python3 -m rospkg.os_detect
  OS Name:     ubuntu
  OS Version:  18.04
  OS Codename: bionic
  ```

This patch should be upstreamed soon.


robwoolley:
Thanks @vermaete 

Is this something we can push upstream to python3-rospkg first and then backport?

I see that there are some comments about distro.id VS disto.name.  This is useful.  At the next meeting let's chat about what OpenEmbedded does for LSB to make sure that we get the mapping right.

PR #1799, opened 2026-08-09
[master]{common}{ros2}Vermaete/duplicate

vermaete:
Remove some duplicated recipes.
Mainly because new version of it in other layers will be used anyhow.


vermaete:
@robwoolley some cleanup in meta-ros-common

